Issue #435, opened 2023-08-22
Slider prefab should have and AudioSource pre-attached


david-c-kline:
## Describe the bug

The current implementation of the MRTK Slider controls (canvas based) have a SliderSounds script attached which dynamically adds an AudioSource component (it does it twice, in fact, once for each type of sound).

To facilitate assigning the sounds to AudioMixer channels (for volume attenuation, muting, etc.) these AudioSource object(s) should be  pre-attached or, at the very least, the following fields should be serialized.

        private AudioSource grabReleaseAudioSource = null;
        private AudioSource passNotchAudioSource = null;

## Expected behavior

The ability to pre-configure the AudioMixerGroup field of the AudioSource component without needing to write runtime scripting.

## Your setup (please complete the following information)

- Unity Version 2021.3.16f1
- MRTK Version pre.17

## Target platform (please complete the following information)

- All

PR #436, opened 2023-08-22
Enable developer customers to access AudioSource at development time

david-c-kline:
This change serializes the grabReleaseAudioSource and passNotchAudioSource AudioSource fields, so that developer customers can configure options such as AudioMixerGroup, spatialization and play on awake.

Fixes: #435 

Verified in Unity editor by:

- Adding a single AudioSource to the slider in the HandInteractionExamplesScene
- Assigning the manually attached AudioSource to the Grab/Release and PassNotch fields (newly serialized)
- Playing the scene in the editor and moving the slider
- Noting
  - No additional AudioSource objects were added to the slider
  - Audio plays correctly
  - The expected AudioSource parameters are updated while interacting with the slider

### DIFF
--- a/org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
+++ b/org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
@@ -113,9 +113,27 @@ namespace MixedReality.Toolkit.UX
         [Tooltip("The source of the grab and release sounds")]
         private AudioSource grabReleaseAudioSource = null;
 
+        /// <summary>
+        /// Gets or sets the AudioSource used to play grab and release sounds.
+        /// </summary>
+        public AudioSource GrabReleaseAudioSource
+        {
+            get => grabReleaseAudioSource;
+            set => grabReleaseAudioSource = value;
+        }
+
         [SerializeField]
         [Tooltip("The source of the pass notch sounds")]
         private AudioSource passNotchAudioSource = null;
+
+        /// <summary>
+        /// Gets or sets the AudioSource used to play sound when the user moves the slider thumb past a notch.
+        /// </summary>
+        public AudioSource PassNotchAudioSource
+        {
+            get => passNotchAudioSource;
+            set => passNotchAudioSource = value;
+        }
 
         #region Private members
         private Slider slider;
--- a/org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
+++ b/org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
@@ -109,6 +109,13 @@ namespace MixedReality.Toolkit.UX
         /// </summary>
         public float MinSecondsBetweenTicks { get => minSecondsBetweenTicks; set => minSecondsBetweenTicks = value; }
 
+        [SerializeField]
+        [Tooltip("The source of the grab and release sounds")]
+        private AudioSource grabReleaseAudioSource = null;
+
+        [SerializeField]
+        [Tooltip("The source of the pass notch sounds")]
+        private AudioSource passNotchAudioSource = null;
 
         #region Private members
         private Slider slider;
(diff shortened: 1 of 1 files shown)


org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
```diff
@@ -109,6 +109,13 @@ public class SliderSounds : MonoBehaviour
         /// </summary>
         public float MinSecondsBetweenTicks { get => minSecondsBetweenTicks; set => minSecondsBetweenTicks = value; }
 
+        [SerializeField]
+        [Tooltip("The source of the grab and release sounds")]
+        private AudioSource grabReleaseAudioSource = null;
```
AMollis, line 114: I have some questions about the original design of the sounds here.

Why they need for two audio sources?   What case would require two or more?

AMollis: Also...this code in Start() could be improved, assuming using a single AudioSource works. If both `grabReleaseAudioSource` and `grabRelease` are null, create a single audio source instead of two.

May be change 

```
            if (grabReleaseAudioSource == null)
            {
                grabReleaseAudioSource = gameObject.AddComponent<AudioSource>();
            }
            if (passNotchAudioSource == null)
            {
                passNotchAudioSource = gameObject.AddComponent<AudioSource>();
            }
```

to

```
            if (grabReleaseAudioSource == null && passNotchAudioSource == null)
            {
                grabReleaseAudioSource = passNotchAudioSource  = gameObject.AddComponent<AudioSource>();
            }
            else if (grabReleaseAudioSource == null)
            {
                grabReleaseAudioSource = passNotchAudioSource ;
            }
            else if (passNotchAudioSource == null)
            {
                passNotchAudioSource = grabReleaseAudioSource;
            }
```

AMollis, line 114: Since you're making this serializable fields.   Do you see benefit in exposing these as properties, so to enable component creation at runtime?

david-c-kline: > I have some questions about the original design of the sounds here.
> 
> Why they need for two audio sources? What case would require two or more?

This is the first time i have looked at this script and I had similar questions. For this change, I strove for the minimum amount of changes.

david-c-kline: I can see doing that. Will update the PR shortly.

david-c-kline: Public accessors added.

david-c-kline: Please let me know if you believe making a larger scale change is needed here. If so, I will investigate / test.

AMollis (approved):
Signing off on changes.

@david-c-kline , can you please look into refactoring to use a single audio source? Looking at the history of the file (from our internal repository) looks like @MaxWang-MS added the two audio sources when they first created the file.

I've opened an issue to track remove the two audio sources.

- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/438

david-c-kline:
> Signing off on changes.
> 
> @david-c-kline , can you please look into refactoring to use a single audio source? Looking at the history of the file (from our internal repository) looks like @MaxWang-MS added the two audio sources when they first created the file.
> 
> I've opened an issue to track refactoring the two audio sources.
> 
> * [[FEATURE REQUEST] Refactor `SlideSounds` to use a single audio source. #438](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/438)

Will do. One thing to note is that using a single audio source will cause the grab/release sounds to have their pitch altered based on the position of the slider thumb (the pass notch code sweeps the frequency of the audio source).

If we are ok with that, I will look into #438, otherwise, we should continue using two audio sources.

PR #464, opened 2023-09-08
Changes to support package releases post GA

shaynie:
Post GA, releases will not be monolithic.  Each package will be versioned and released separately as needed.
The changes in this pull request support the new way of releasing changes. 

Specifically:
  * Update the list of released packages so they don't get tagged with the current preview tag (pre.19)
    * The accessibility and data packages will still get tagged.
    
  * Update all dependencies on other toolkit packages to version "3".  EDIT:  Unity doesn't recognize the naked 3 as a valid semantic version, so leaving the dependencies as 3.0.0
    * If a package (say, core) has new functionality so that it gets a new minor version i.e. 3.1, then packages that depend on core should only update their dependency version to 3.1 if they rely on the new functionality.

  * Change the pack-upm and repackage-for-release scripts to not rewrite dependencies anymore

  * Update the uxcore package to version 3.1 to reflect that it has new functionality per: https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/436

 The CI, PR, and RC builds will continue to build each package, but when a release is necessary only the changed packages will be published.

Please refer to the "[Versioning and Releases](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/main/contributions/versioning-and-releases.md)" document for more information.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -23,10 +23,10 @@
 * @MixedRealityToolkit/mixedrealitytoolkit-unity-maintainers
 
 # Build / CI Pipeline
-/Pipelines/ @shaynie @keveleigh
+/Pipelines/ @shaynie @amollis @keveleigh
 
 # Tooling
-/Tooling/ @shaynie @keveleigh @david-c-kline
+/Tooling/ @shaynie @amollis @keveleigh @david-c-kline
 
 # Feature packages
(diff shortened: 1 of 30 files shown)


AMollis:
It would probably be worth testing the MRFT with the new dependency versions.

AMollis (commented):
Please confirm this change works in Microsoft's Mixed Reality Feature Tool.  On confirmation, I'll sign off :)

AMollis (approved):

Issue #472, opened 2023-09-13
[BUG] The Slider step divisions feature has issues when the min and max values are not 0 and 1


ghazen-ml:
## Describe the bug

The MRTK Slider UX component includes a feature to cause snapping to a specific number of subdivisions. E.g. if set to 5, there will be 5 snap points along the slider's length. However if you set the Min and Max values to anything but 0 and 1, it seems that the slider will get stuck.

## To reproduce

1. Go to the Canvas Tearsheet scene in the dev templates
2. Set the Max value of one of the sliders to 20, and enable 5 slider step divisions
3. Launch the scene
4. Try using the slider. Notice that the slider handle is stuck at 0 and won't move to other values.

## Expected behavior

The slider should support movement side to side to the 5 step divisions

## Screenshots

![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/65045943/9adc1ba4-c38e-46af-8523-5d759edd1116)

## Your setup (please complete the following information)

- Unity Version 2021.3.21f1
- MRTK Version or Commit e3002c17a9294c223ce7ab7296b732775bf92d9c

## Target platform (please complete the following information)

- Unity Editor

marlenaklein-msft:
Thanks for bug. This should be resolved in #479

PR #479, opened 2023-09-14
Fix Slider Snapped Step Positions with Non-Default Range

marlenaklein-msft:
Switching the math around so that non-normalized value is passed into SnapSliderToStepPosition.


marlenaklein-msft:
Fixes #472

shaynie (approved):

Issue #466, opened 2023-09-11
Document how critical bugs should be resolved and released


AMollis:
## Describe the issue

In the unlikely event of a critical bug, what process should the Project Maintainers follow to address the issue in a timely manner?  We should have clear guidelines outlining the escalation process for Project Maintainers.  

## Feature area

Project Management

## Existing doc link

It probably makes sense to append this documentation with the critical bug process.

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/main/contributions/versioning-and-releases.md

shaynie:
@shaynie  or @AMollis  to send emergency email to discuss with maintainers when critical bugs arise.  @AMollis to set up mail group.

PR #483, opened 2023-09-21
Adding guidance for addressing critical issues

AMollis:
Publicly documenting how critical issues are handled by Project Maintainers.

## Issues

- Fixes: https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/466

### DIFF
--- a/contributions/opening-and-assessing-issues.md
+++ b/contributions/opening-and-assessing-issues.md
@@ -115,7 +115,7 @@ In place of assigning a milestone, consider assigning issues to a GitHub [projec
 
 ## Critical priority issues
 
-Such a priority must be used sparingly and only applies to released package issues that severely impact a large number of users. For example, an issue that results in data loss for all users can be considered critical, while an issue causing compile errors for some developers does not warrant a critical priority.
+Such a priority must be used sparingly and only applies to released package issues that severely impact a large number of users. For an issue to be deemed critical it must be sever enough to require the immediate attention of the Project Maintainers. Security and data loss bugs, while unlikely in MRTK, are examples of issues that may warrant critical status.
 
 High priority issues that aren't deemed critical should be addressed during scheduled Project Maintainer meetings. These meetings typically occur every two weeks.
 
--- a/contributions/opening-and-assessing-issues.md
+++ b/contributions/opening-and-assessing-issues.md
@@ -117,4 +117,6 @@ In place of assigning a milestone, consider assigning issues to a GitHub [projec
 
 Such a priority must be used sparingly and only applies to released package issues that severely impact a large number of users. For example, an issue that results in data loss for all users can be considered critical, while an issue causing compile errors for some developers does not warrant a critical priority.
 
+High priority issues that aren't deemed critical should be addressed during scheduled Project Maintainer meetings. These meetings typically occur every two weeks.
+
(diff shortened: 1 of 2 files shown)


marlenaklein-msft (approved):

shaynie (approved):

contributions/opening-and-assessing-issues.md
```diff
@@ -109,6 +109,12 @@ GitHub [milestones](https://docs.github.com/en/issues/using-labels-and-milestone
...
 Before a new milestone can be created in the repository, the new milestone must be approved by the project Maintainers, following the decision making rules in the [GOVERNANCE.md](../GOVERNANCE.md) file.
 
-### GitHub project assignments
+## GitHub project assignments
 
 In place of assigning a milestone, consider assigning issues to a GitHub [project](https://docs.github.com/en/issues/planning-and-tracking-with-projects/learning-about-projects/about-projects). An issue can have multiple project assignments, and each Project Maintainer can manage their own GitHub projects and the priorities within those projects.
+
+## Critical priority issues
+
+Such a priority must be used sparingly and only applies to released package issues that severely impact a large number of users. For example, an issue that results in data loss for all users can be considered critical, while an issue causing compile errors for some developers does not warrant a critical priority.
```
keveleigh, line 118: Strongly disagree that compilation errors for _any_ developer should be considered "not [...] critical". Intentionally locking out any portion of the developer base without a patch release is not a stance we should take.

keveleigh, line 118: Can we replace these with examples relevant to this project? I'm not sure anybody's ever had data loss from MRTK...

keveleigh (changes requested):

contributions/opening-and-assessing-issues.md
AMollis: I struggle with an example here.   However, I don't foresee a critical issue in our future.   The support team here at Microsoft wanted to make sure we have a plan in case a security issue should arise.   

Happy for you to add a better example here.

AMollis: We discussed at the maintainer sync. Every two weeks is the normal cadence of updates.  So other high priority issue can be discussed during these normal sync meetings.

Critical issues are meant to be addressed outside those two weeks.   I'll add some text to clarify this.

AMollis:
@keveleigh  Some changes for you :)

shaynie (approved):

contributions/opening-and-assessing-issues.md
keveleigh: Ah, good to know that there was missing context. Thanks

keveleigh: > That's sort of the point :) Critical type things should be rare/unlikely.

I hope I'm misunderstanding this statement, but (the way I'm reading it) I'm not sure I agree. Aspirationally, sure, we should make sure we don't push out critical problems often. But we shouldn't _define_ critical events around things that are unlikely to happen. Criticality should be based on severity, not frequency.

The easiest example, based around the existing example (assuming at this point that it's not changing regardless of feedback), could be something like:

> For example, an issue that results in **_compile errors_** for all users can be considered critical, while an issue causing compile errors for some developers does not warrant a critical priority.

Hopefully you agree that an unbuildable MRTK is a critical issue 😉

keveleigh: Oh, I see that the examples were removed in the latest iteration. Fair enough!

Issue #493, opened 2023-09-27
Check to see if UX Components (non-canvas) needs a version bump for slider prefabs


shaynie:

shaynie:
@david-c-kline - check this package as well

AMollis:
@david-c-kline  is this done?

Issue #438, opened 2023-08-23
[FEATURE REQUEST] Refactor `SliderSounds` to use a single audio source.


AMollis:
## Overview

The component at [org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/main/org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs) has two audio sources, `passNotchAudioSource` and `grabReleaseAudioSource`.  We don't understand why there are two audio sources.

Both `passNotchAudioSource` and `grabReleaseAudioSource` were private non-serialized fields.  However, with the following PR, these fields have been exposed via the inspector window and public properties. 

- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/436

Marking to this as a "Release Blocker" to avoid shipping an update to "UX Core" without first understanding this issue. To avoid future deprecation, if two audio sources aren't needed, we should not spin a release with these non-private audio sources.

## Tasks

- [ ] Determine if two audio sources are in fact needed.
- [ ] If two audio sources aren't needed, refactor code so that there is only one.
- [ ] If two audio source are needed, close this issue.

## Additional context

Looking at the history of the file (from our internal repository) looks like @MaxWang-MS added the two audio sources when they first created the file.

david-c-kline:
One thing to note is that using a single audio source will cause the grab/release sounds to have their pitch altered based on the position of the slider thumb (the pass notch code sweeps the frequency of the audio source).

If we are ok with that, we can unify the sources, otherwise, we should continue using two.

david-c-kline:
@AMollis, @keveleigh, @shaynie, @ghazen-ml, @whebertML Thoughts?

whebertML:
@david-c-kline @AMollis Perhaps an alternative to exposing that two AudioSources are used internally so as to avoid possible future deprecation if that changes, just expose one serialized AudioSource that the dev can tweak, and internally make a second one for pass notch use and duplicate the settings (like audio mixer group) and keep the benefit of the pitch changes for pass notch?

AMollis:
> One thing to note is that using a single audio source will cause the grab/release sounds to have their pitch altered based on the position of the slider thumb (the pass notch code sweeps the frequency of the audio source).
> 
> If we are ok with that, we can unify the sources, otherwise, we should continue using two.

[...]

PR #495, opened 2023-09-28
Update SliderSounds to use one AudioSource instead of two

david-c-kline:
This change updates SliderSounds.cs (UXCore) to use one AudioSource object for both gab/release and notch sounds as well as exposing the new field in the inspector and as a property. This is a change from the previous implementation where the developer did NOT provide AudioSources (automatically created by SliderSounds). This new implementation will auto-create the necessary object if not specified by the developer.

Since the notch sound lerps the pitch of the clip. this change caches and restores pitch shift as needed. It has been audibly and visually verified in the Unity Editor's play mode.

To complete this change, the canvas and non-canvas base prefabs have been updated to contain a reference to an explicit AudioSource object so that developers can customize properties (ex: mixer channel, spatialization settings, etc.).

The following package versions (and relevant dependencies have been updated)
UXCore: 3.1.0
UXComponents: 3.1.0 (depends upon UXCore 3.1.0)
UXComponents.NonCanvas: 3.1.0 (depends upon UXCore 3.1.0)

Fixes: #438
Fixes: #493


org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
```diff
@@ -102,6 +102,7 @@ MonoBehaviour:
   _Iridescent_Map_: {fileID: 0}
   _Frequency_: 1.89
   _Vertical_Offset_: 0
+  _Orthographic_Distance_: 400
```
marlenaklein-msft, line 105: are all of these prefab changes desired?

org.mixedrealitytoolkit.uxcomponents.noncanvas/Slider/NonCanvasSliderBase.prefab
```diff
@@ -154,8 +154,8 @@ Transform:
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 2624031095332604921}
   m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
-  m_LocalPosition: {x: 0, y: 0, z: 0.000000012922101}
-  m_LocalScale: {x: 0.099999905, y: 0.002, z: 1}
+  m_LocalPosition: {x: -9.313226e-10, y: 0, z: 0}
```
AMollis, line 157: why the position change?

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
```diff
@@ -248,7 +251,7 @@ RectTransform:
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_AnchorMin: {x: 0.5, y: 0.5}
   m_AnchorMax: {x: 0.5, y: 0.5}
-  m_AnchoredPosition: {x: 0, y: 0}
+  m_AnchoredPosition: {x: -0.000017166138, y: -0.000002861023}
```
AMollis, line 254: Why this change?

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
```diff
@@ -151,13 +136,9 @@ public AudioSource PassNotchAudioSource
...
         {
-            if (grabReleaseAudioSource == null)
-            {
-                grabReleaseAudioSource = gameObject.AddComponent<AudioSource>();
-            }
-            if (passNotchAudioSource == null)
+            if (audioSource == null)
             {
-                passNotchAudioSource = gameObject.AddComponent<AudioSource>();
+                audioSource = gameObject.GetComponent<AudioSource>();
```
AMollis, line 141: We should still create a AudioSource if missing, to maintain old behavior.   It's feasible apps could be using their own slider prefab, relying on the auto creation of the audio source.

```
if (audioSource == null)
{
    audioSource = gameObject.GetComponent<AudioSource>();
    if (audioSource == null)
    {
        audioSource = gameObject.AddComponent<AudioSource>();
   }
}

**Review by AMollis — DISMISSED** [review]
See other comments for change requests

File: org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs [hunk 5]
```diff
@@ -171,45 +152,51 @@ private void Start()
...
+        private static float NeutralPitchShift = 1f;
+        private float passNotchPitchShift = NeutralPitchShift;
+ 
         private void OnValueUpdated(SliderEventData eventData)
         {
-            if (!(playSoundsOnlyOnInteract && !isInteracting) && playTickSounds && passNotchAudioSource != null && passNotchSound != null)
+
+            if (!(playSoundsOnlyOnInteract && !isInteracting) && playTickSounds && AudioSource != null && passNotchSound != null)
             {
+                AudioSource.pitch = passNotchPitchShift;
```
whebertML, line 163: Since AudioSource.pitch is recalculated every time (line 169) before playing the passNotchSound, what is the purpose of storing the passNotchPitchShift and setting it here?

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
david-c-kline: these MIGHT have happened because I was using Unity 2021.3.30f1. There are a number of packages that do not load on 2021.3.16f1 that are referenced in the sample project manifest.

org.mixedrealitytoolkit.uxcomponents.noncanvas/Slider/NonCanvasSliderBase.prefab
david-c-kline: zero clue. i moved _nothing_ in the prefab. only added the AudioSouurce component. This looks like floating point error on the reserialization

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
david-c-kline: again, i changed nothing explicitly. Unity did this while reserializing

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
david-c-kline: will make the change.

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
david-c-kline: i was being cautious :smile:

i can remove

david-c-kline:
on the unity prefab serialization changes. other than adding the AudioSource component, there were no explicit changes made during the creation of this PR.

if required, I can work on a different machine and install the exact matching Unity version that is specified in the template's project settings.

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
AMollis: Can you please revert these changes?

org.mixedrealitytoolkit.uxcomponents.noncanvas/Slider/NonCanvasSliderBase.prefab
AMollis: please revert

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
keveleigh: Might be worth checking for the field being null _and_ checking `TryGetComponent` in the `if`, then adding if it's not found through either of those paths.

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
```diff
@@ -110,29 +108,16 @@ public class SliderSounds : MonoBehaviour
...
-        private AudioSource passNotchAudioSource = null;
+        [Tooltip("The AudioSource to be used when playing slider sounds.")]
+        private AudioSource audioSource = null;
 
         /// <summary>
-        /// Gets or sets the AudioSource used to play sound when the user moves the slider thumb past a notch.
+        /// Gets or sets the AudioSource used to play slider sounds.
         /// </summary>
-        public AudioSource PassNotchAudioSource
+        public AudioSource AudioSource
```
keveleigh, line 117: nit: do we want to move to using the `[field: SerializeField]` pattern on an auto-property here?

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
```diff
@@ -110,29 +108,16 @@ public class SliderSounds : MonoBehaviour
...
-        {
-            get => grabReleaseAudioSource;
-            set => grabReleaseAudioSource = value;
-        }
-
-        [SerializeField]
-        [Tooltip("The source of the pass notch sounds")]
-        private AudioSource passNotchAudioSource = null;
+        [Tooltip("The AudioSource to be used when playing slider sounds.")]
+        private AudioSource audioSource = null;
```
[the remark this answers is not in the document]
whebertML: I think @david-c-kline previously added serialization of these two fields in a very recent earlier commit, https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/1d612518fd1c5de34962d443114f6edcdd4324c7, but that did not make it into the GA release, I don't believe.  So the serialization of these isn't yet part of an official release, so might not be an issue?

org.mixedrealitytoolkit.uxcomponents.noncanvas/Slider/NonCanvasSliderBase.prefab
david-c-kline: i tried reverting and Unity 2021.3.21f` (matching version for the template project) reserialized this in as well.

the only thing i can think is to NOT make the prefab changes

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
david-c-kline: 2021.3.21f1 (matching version) also added these entries. it appears that Unity changed serialization. Unfortunately, it is not clear what version caused these to be added (and the project does not load with earlier versions due to package incompatibilities)

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
david-c-kline: See above comment (Unity did this during their serialization). I can _not_ update the prefabs, but that eliminates part of the benefit of this change.

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
marlenaklein-msft: ah okay, makes sense. usually for prefab changes I will just manually unstage anything extraneous that Unity does / manually stage the changes I want

org.mixedrealitytoolkit.uxcomponents.noncanvas/Slider/NonCanvasSliderBase.prefab
keveleigh: I'm able to repro these changes locally using Unity 2021.3.21f1 (down to the last decimal point, which is actually kinda impressive haha) by doing nothing except adding the AudioSource component.

image

This seems like some Unity-internal serializer update, so I'm pretty sure it's unavoidable and we should take the changes now or we'll be plagued with them changing locally every time a change is made to these prefabs.

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
marlenaklein-msft: (also hopefully the project will load with earlier versions of 2021 after the recent changes to avoid compile errors)

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
david-c-kline: that is the plan :)

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
david-c-kline: the project is not loadable in previous versions due to an updated Unity package (not sure which one) that is not compatible with 16f1, for example.

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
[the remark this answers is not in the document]
david-c-kline: @whebertML, correct!

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
keveleigh: Yeah! For example:

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/125e8add28d5837150e0a177483a346a3999a486/org.mixedrealitytoolkit.core/Interactables/StatefulInteractable.cs#L95-L96

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
marlenaklein-msft: I was able to run the samples in 16f1 after this change https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/484, are there other errors in 16f1?

david-c-kline: i'l double check. there _had_ been a package issue in the past.

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
david-c-kline: as long as this pattern matches the rest of the file, happy to make the change :)

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
david-c-kline: @marlenaklein-msft, you are correct. the template project opens successfully in 16f1 :smile:

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
david-c-kline: decided to minimize the change to this file and maintain consistency with the existing code pattern

org.mixedrealitytoolkit.uxcomponents/Slider/CanvasSlider.prefab
david-c-kline: resolving: this change is a result of reserializing with 2021.3. Any future change will generate these changes. Let's keep them now, so we do not have this same discussion at that time.  😄

org.mixedrealitytoolkit.uxcomponents.noncanvas/Slider/NonCanvasSliderBase.prefab
david-c-kline: resolving: per @keveleigh's repro and that this is a Unity 2021.3 reserialization, i believe we should accept this change so that these questions do not arise for any future change to this prefab

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
david-c-kline: change made.

david-c-kline:
@MixedRealityToolkit/mixedrealitytoolkit-unity-maintainers 

This PR should be ready for merging.

marlenaklein-msft (approved):

AMollis (approved):

Issue #66, opened 2023-08-20
Update MRTK3 to use Unity XR Hands package


IssueSyncBot:
### Original issue opened by:
<img class="avatar avatar-user" src="https://avatars.githubusercontent.com/u/36461279?v=4" width="20" height="20" style="border-radius: 50%" alt="@AMollis"> @AMollis
___

## Describe the problem

With Unity's [XR Hands](https://docs.unity3d.com/Packages/com.unity.xr.hands@1.1/manual/index.html) package, Unity now exposes articulates hand tracking data.  MRTK3 should be update to the utilize this package.

The easiest route to create a MRTK3 `HandsSubsystem` that wraps [XR Hands](https://docs.unity3d.com/Packages/com.unity.xr.hands@1.1/manual/index.html) data.

## Tasks

- [x]  Create a MRTK3 `HandsSubsystem` that wraps [XR Hands](https://docs.unity3d.com/Packages/com.unity.xr.hands@1.1/manual/index.html) `XRHandSubsystem`.
- [x] Update MRTK3 samples to use new MRTK3 `HandsSubsystem`
- [ ] Update MRTK3 docs

___
### ISSUE MIGRATION

   Issue migrated from: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/11680

shaynie:
@keveleigh  has this in progress.  Review PR #504

shaynie:
Kurtis will do a doc PR about how to enable

Consider making this the default in the next major version

PR #500, opened 2023-10-02
Refactor out HandsProvider<T>

keveleigh:
First step of https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/66, to reduce the boilerplate code needed to spin up a new hands provider.

Also refactors out `DefaultHandJointRadius`, as `0.005f` was used in a couple places.

Also, some assorted formatting and slight perf updates.

### DIFF
--- a/org.mixedrealitytoolkit.input/Subsystems/Hands/HandsProvider.cs
+++ b/org.mixedrealitytoolkit.input/Subsystems/Hands/HandsProvider.cs
@@ -13,9 +13,11 @@ namespace MixedReality.Toolkit.Input
 {
     /// <summary>
     /// Provides a generic, base hands provider for subsystem use.
+    /// </summary>
+    /// <remarks>
     /// Extends the <see cref="MixedReality.Toolkit.Subsystems.HandsSubsystem.Provider">Provider</see> class and
     /// obtains hand joint poses from a <see cref="MixedReality.Toolkit.Input.HandDataContainer">HandDataContainer</see> class.
-    /// </summary>
+    /// </remarks>
     /// <typeparam name="T">
     /// The type of <see cref="MixedReality.Toolkit.Input.HandDataContainer">HandDataContainer</see> to query hand data from.
     /// </typeparam>
--- a/org.mixedrealitytoolkit.input/Subsystems/Hands/HandsProvider.cs
+++ b/org.mixedrealitytoolkit.input/Subsystems/Hands/HandsProvider.cs
@@ -1,0 +1,70 @@
+using MixedReality.Toolkit.Subsystems;
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.Scripting;
+using UnityEngine.XR;
+
+namespace MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Provides a generic, base hands provider for subsystem use.
+    /// Extends the <see cref="MixedReality.Toolkit.Subsystems.HandsSubsystem.Provider">Provider</see> class, and
+    /// obtains hand joint poses from a <see cref="MixedReality.Toolkit.Input.HandDataContainer">HandDataContainer</see> class.
+    /// </summary>
+    /// <typeparam name="T">
+    /// The type of <see cref="MixedReality.Toolkit.Input.HandDataContainer">HandDataContainer</see> to query hand data from.
+    /// </typeparam>
+    [Preserve]
+    internal class HandsProvider<T> : HandsSubsystem.Provider where T : HandDataContainer
+    {
+        private Dictionary<XRNode, T> hands = null;
+
+        /// <inheritdoc/>
+        public override void Start()
+        {
+            base.Start();
+
+            hands ??= new Dictionary<XRNode, T>
+            {
+                { XRNode.LeftHand, Activator.CreateInstance(typeof(T), XRNode.LeftHand) as T },
+                { XRNode.RightHand, Activator.CreateInstance(typeof(T), XRNode.RightHand) as T }
+            };
+
+            InputSystem.onBeforeUpdate += ResetHands;
+        }
+
+        /// <inheritdoc/>
+        public override void Stop()
+        {
+            ResetHands();
+            InputSystem.onBeforeUpdate -= ResetHands;
+            base.Stop();
+        }
+
+        private void ResetHands()
+        {
+            hands[XRNode.LeftHand].Reset();
+            hands[XRNode.RightHand].Reset();
+        }
+
+        #region IHandsSubsystem implementation
+
+        /// <inheritdoc/>
+        public override bool TryGetEntireHand(XRNode handNode, out IReadOnlyList<HandJointPose> jointPoses)
+        {
+            Debug.Assert(handNode == XRNode.LeftHand || handNode == XRNode.RightHand, "Non-hand XRNode used in TryGetEntireHand query.");
+            return hands[handNode].TryGetEntireHand(out jointPoses);
+        }
+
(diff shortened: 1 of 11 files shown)


org.mixedrealitytoolkit.input/Subsystems/Hands/OpenXRHandsSubsystem.cs
```diff
@@ -25,7 +24,7 @@ namespace MixedReality.Toolkit.Input
         Name = "org.mixedrealitytoolkit.openxrhands",
         DisplayName = "Subsystem for OpenXR Hands API",
         Author = "Mixed Reality Toolkit Contributors",
-        ProviderType = typeof(OpenXRProvider),
+        ProviderType = typeof(HandsProvider<OpenXRHandContainer>),
```
david-c-kline, line 27: I may be missing something (bad eyes?) ... where is OpenXRHandContainer defined?

org.mixedrealitytoolkit.input/Subsystems/Hands/XRSDKHandsSubsystem.cs
```diff
@@ -22,7 +21,7 @@ namespace MixedReality.Toolkit.Input
         Name = "org.mixedrealitytoolkit.xrsdkhands",
         DisplayName = "Subsystem for XRSDK Hands API",
         Author = "Mixed Reality Toolkit Contributors",
-        ProviderType = typeof(XRSDKProvider),
+        ProviderType = typeof(HandsProvider<XRSDKHandContainer>),
```
david-c-kline, line 24: Similar question re: XRSDKHandContainer

david-c-kline (commented):
Please be sure to increment the version number in the core's package.json file and any required dependencies (ex: input)

org.mixedrealitytoolkit.input/Subsystems/Hands/OpenXRHandsSubsystem.cs
keveleigh: They're existing private classes defined in their corresponding subsystem files:

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/62dca36d8fdae7c0c17910f10b1eca2f22ed3e70/org.mixedrealitytoolkit.input/Subsystems/Hands/OpenXRHandsSubsystem.cs#L54

org.mixedrealitytoolkit.input/Subsystems/Hands/XRSDKHandsSubsystem.cs
keveleigh: They're existing private classes defined in their corresponding subsystem files.

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/62dca36d8fdae7c0c17910f10b1eca2f22ed3e70/org.mixedrealitytoolkit.input/Subsystems/Hands/XRSDKHandsSubsystem.cs#L49

keveleigh:
> Please be sure to increment the version number in the core's package.json file and any required dependencies (ex: input)

Core had already been incremented to 3.0.1 in #485 (which hasn't shipped yet), so I only incremented Input to 3.0.1 (along with its Core dependency version).

Since the changes in this PR are entirely internal, I don't think it falls under the category of a minor version bump. Once I complete #66, it will though!

david-c-kline (approved):

org.mixedrealitytoolkit.input/Subsystems/Hands/HandsProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.Scripting;
+using UnityEngine.XR;
+
+namespace MixedReality.Toolkit.Input
+{
+    /// <summary>
```
AMollis, line 14: Can you please add the second sentence to the "remarks" tag.   On learn.microsoft.com, the summary bit is shown in a table alongside all other types in the assembly.  So having a shorter summary is preferred in that view.

```
    /// <summary>
    /// Provides a generic, base hands provider for subsystem use.
    /// </summary>
    /// <remarks>
    /// Extends the <see cref="MixedReality.Toolkit.Subsystems.HandsSubsystem.Provider">Provider</see> class and
    /// obtains hand joint poses from a <see cref="MixedReality.Toolkit.Input.HandDataContainer">HandDataContainer</see> class.
    /// </remarks>
```

org.mixedrealitytoolkit.input/package.json
```diff
@@ -16,7 +16,7 @@
   "unity": "2021.3",
   "dependencies": {
     "com.microsoft.mrtk.graphicstools.unity": "0.5.12",
-    "org.mixedrealitytoolkit.core": "3.0.0",
+    "org.mixedrealitytoolkit.core": "3.0.1",
```
AMollis, line 19: Why force an upgrade to a patch version?  Scott and I have been wanted to only require updates if the package requires a feature change, or if there is a breaking change to a dependency.   Otherwise, we should give apps the option of when to upgrade.

AMollis (changes requested):

org.mixedrealitytoolkit.input/package.json
keveleigh: There's a new `internal` field in Core that Input relies on:

image

Because there's no change to the public API surface nor substantial new private functionality in Core, it doesn't match up with a minor version bump. Therefore, the dependency was increased to this patch version.

org.mixedrealitytoolkit.input/Subsystems/Hands/HandsProvider.cs
keveleigh: I made this change and will note it for the future, but I didn't think `internal` classes ended up on learn.microsoft.com.

AMollis: Doh it's internal...sorry about that comment.

org.mixedrealitytoolkit.input/package.json
AMollis: Ack...I'm even more confused.

1. We already started the release for 3.0.1.   So you should update to MRTK Core to 3.0.2
2. How can outside assembly reference an internal field from a different assembly.  What's the point of marking something internal then?   I see that it's working, but I don't understand.

AMollis (changes requested):

org.mixedrealitytoolkit.input/package.json
keveleigh: 1. Can do! (and done!)
2. We use the `InternalsVisibleTo` tag in an AssemblyInfo.cs file. We've mainly used this pattern to define something in Core that we want to be well-defined (or want to reserve the right to change the implementation of) across all packages but not expose publicly for one reason or another. From a quick check, it looks like the two classes that are accessed in Input (from Core) this way are HandsUtils.cs and PoseUtilities.cs.

    https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/e2c3e55981876487e5e57c4f495f5aedcf8b52bf/org.mixedrealitytoolkit.core/AssemblyInfo.cs#L10

AMollis: Ahh...InternalsVisibleTo  ... You learn something new everyday.  Thanks for the explanation.

AMollis (approved):

org.mixedrealitytoolkit.input/package.json
keveleigh: Happy to help!

david-c-kline (approved):

PR #504, opened 2023-10-06
Unity XR Hands provider

keveleigh:
Next phase of #66: actually creating the hands provider wrapping Unity's XR Hands API.

1. Creates the `HandsSubsystem` implementation wrapping Unity's XR Hands API.
2. As a result, adds Unity's XR Hands package as a hard dependency to the Input package.
3. DOES NOT yet change any MRTK profiles or sample scenes. That will come in the next PR.

Does not complete https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/66 but is a major step along to completion.

### DIFF
--- a/org.mixedrealitytoolkit.input/Subsystems/Hands/UnityHandsSubsystem.cs
+++ b/org.mixedrealitytoolkit.input/Subsystems/Hands/UnityHandsSubsystem.cs
@@ -13,8 +13,8 @@ namespace MixedReality.Toolkit.Input
 {
     [Preserve]
     [MRTKSubsystem(
-        Name = "org.mixedrealitytoolkit.unityhands",
-        DisplayName = "Subsystem for Unity Hands API",
+        Name = "org.mixedrealitytoolkit.unityxrhands",
+        DisplayName = "Subsystem for Unity XR Hands API",
         Author = "Mixed Reality Toolkit Contributors",
         ProviderType = typeof(HandsProvider<UnityHandContainer>),
         SubsystemTypeOverride = typeof(UnityHandsSubsystem),
--- a/org.mixedrealitytoolkit.input/Subsystems/Hands/UnityHandsSubsystem.cs
+++ b/org.mixedrealitytoolkit.input/Subsystems/Hands/UnityHandsSubsystem.cs
@@ -118,7 +118,7 @@ namespace MixedReality.Toolkit.Input
 
             /// <summary/>
             /// For a certain hand, query every Bone in the hand, and write all results to the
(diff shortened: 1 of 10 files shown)


org.mixedrealitytoolkit.input/Subsystems/Hands/UnityHandsSubsystem.cs
```diff
@@ -0,0 +1,259 @@
...
+namespace MixedReality.Toolkit.Input
+{
+    [Preserve]
+    [MRTKSubsystem(
+        Name = "org.mixedrealitytoolkit.unityxrhands",
+        DisplayName = "Subsystem for Unity XR Hands API",
+        Author = "Mixed Reality Toolkit Contributors",
+        ProviderType = typeof(HandsProvider<UnityHandContainer>),
+        SubsystemTypeOverride = typeof(UnityHandsSubsystem))]
+    public class UnityHandsSubsystem : HandsSubsystem
```
AMollis, line 21: Please add XML summary block on public types.

AMollis, line 21: Are there any unit tests we can add for this?  We should try to add Unit tests for new components if possible

AMollis (changes requested):

org.mixedrealitytoolkit.input/Subsystems/Hands/UnityHandsSubsystem.cs
AMollis: Probably doesn't make sense ..... but thought I'd ask

keveleigh: Ahh yeah, it's a good thought! This one is harder imo since it's specifically driven by a Unity API. I _think_ it would be possible to create a custom `Unity​Engine.​XR.​Hands.XRHandSubsystem` which is driven by test data to test this flow, but that might be more trouble than the value gained.

keveleigh: Updated!

whebertML (commented):
Can't test XR Hands on our device yet, but all looks pretty good!

AMollis (dismissed):

AMollis (approved):

dahaideyu:
@AMollis  Could you give me simple introduce about  HandsSubsystem ,how to use, for example : how to get click position when your are click, and if your  hand ray  line point position

Issue #502, opened 2023-10-04
[BUG] Add More IfDefs To Get Rid of Warnings in 2023


marlenaklein-msft:
## Describe the bug

In 2023, there are a ton of warnings about various deprecated FindObject functions. Need to replace with the more efficient new versions, similar to this https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/484.

## Screenshots
![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/125413876/e554cdc0-98ba-41d9-a3c0-ab9c8f81695d)

## Your setup (please complete the following information)

- Unity 2023.2

AMollis:
We should fix this :)

AMollis:
@marlenaklein-msft please add the "package: *" labels for all impacted packages.

marlenaklein-msft:
Added. (mostly these are unit tests)

PR #506, opened 2023-10-09
Add Precompile Statements to Prevent Warnings in 2023

marlenaklein-msft:
Adds ifdefs to avoid calling deprecated FindObject functions.

Resolves [this issue.](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/502)

### DIFF
--- a/UnityProjects/MRTKDevTemplate/Assets/Scripts/SampleSceneHandMenu.cs
+++ b/UnityProjects/MRTKDevTemplate/Assets/Scripts/SampleSceneHandMenu.cs
@@ -54,7 +54,11 @@ namespace MixedReality.Toolkit.Examples.Demos
             previousSceneButton.enabled = IsSceneValid(SceneManager.GetActiveScene().buildIndex - 1);
             nextSceneButton.enabled = IsSceneValid(SceneManager.GetActiveScene().buildIndex + 1);
 
+#if UNITY_2022_NEWER
+            SimpleProfiler profiler = Object.FindFirstObjectOfType<SimpleProfiler>(true);
+#else
             SimpleProfiler profiler = Object.FindObjectOfType<SimpleProfiler>(true);
+#endif
             if (profiler != null)
             {
                 profilerObject = profiler.gameObject;
--- a/UnityProjects/MRTKDevTemplate/Assets/Scripts/SampleSceneHandMenu.cs
+++ b/UnityProjects/MRTKDevTemplate/Assets/Scripts/SampleSceneHandMenu.cs
@@ -54,8 +54,8 @@ namespace MixedReality.Toolkit.Examples.Demos
             previousSceneButton.enabled = IsSceneValid(SceneManager.GetActiveScene().buildIndex - 1);
             nextSceneButton.enabled = IsSceneValid(SceneManager.GetActiveScene().buildIndex + 1);
 
-#if UNITY_2022_NEWER
-            SimpleProfiler profiler = Object.FindFirstObjectOfType<SimpleProfiler>(true);
+#if UNITY_2022_1_OR_NEWER
+            SimpleProfiler profiler = Object.FindFirstObjectByType<SimpleProfiler>(FindObjectsInactive.Include);
 #else
             SimpleProfiler profiler = Object.FindObjectOfType<SimpleProfiler>(true);
 #endif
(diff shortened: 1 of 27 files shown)


org.mixedrealitytoolkit.input/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "org.mixedrealitytoolkit.input",
-  "version": "3.0.1-development",
```
keveleigh, line 3: Does this need to be incremented? I don't think we've shipped 3.0.1 of Input yet. (Same question from some of the other package version changes too)

UnityProjects/MRTKDevTemplate/Assets/Scripts/SampleSceneHandMenu.cs
```diff
@@ -54,7 +54,11 @@ private void Awake()
             previousSceneButton.enabled = IsSceneValid(SceneManager.GetActiveScene().buildIndex - 1);
             nextSceneButton.enabled = IsSceneValid(SceneManager.GetActiveScene().buildIndex + 1);
 
+#if UNITY_2022_1_OR_NEWER
+            SimpleProfiler profiler = Object.FindFirstObjectByType<SimpleProfiler>(FindObjectsInactive.Include);
+#else
             SimpleProfiler profiler = Object.FindObjectOfType<SimpleProfiler>(true);
+#endif
```
keveleigh, line 61: I haven't checked exactly how many variants there are throughout, but I wonder if it's worth extracting out static (extension or otherwise) methods to allow the `#if` statements to live in only one place. It would also help reduce the number of asmdef changes.

For example, something like (no guarantees on code correctness here 😅)

```c#
public static class MRTKObject
{
    public static T FindObjectByType<T>(bool includeInactive)
    {
#if UNITY_2022_1_OR_NEWER
        return Object.FindFirstObjectByType<T>(includeInactive ? FindObjectsInactive.Include : FindObjectsInactive.Exclude);
#else
        return Object.FindObjectOfType<T>(includeInactive);
#endif
    }
}

File: UnityProjects/MRTKDevTemplate/Assets/Scripts/SampleSceneHandMenu.cs [hunk 3]
```diff
@@ -54,7 +54,11 @@ private void Awake()
             previousSceneButton.enabled = IsSceneValid(SceneManager.GetActiveScene().buildIndex - 1);
             nextSceneButton.enabled = IsSceneValid(SceneManager.GetActiveScene().buildIndex + 1);
 
+#if UNITY_2022_1_OR_NEWER
```
keveleigh, line 57: Is this method not also `UNITY_2021_3_18_OR_NEWER`? The Unity docs make me believe it might be

https://docs.unity3d.com/2021.3/Documentation/ScriptReference/Object.FindFirstObjectByType.html

org.mixedrealitytoolkit.input/package.json
marlenaklein-msft: I was assuming these changes would go in after the next release... I can change these if needed

UnityProjects/MRTKDevTemplate/Assets/Scripts/SampleSceneHandMenu.cs
marlenaklein-msft: This seems much cleaner! Will refactor

marlenaklein-msft: Then this will be a minor version change update to core

UnityProjects/MRTKDevTemplate/Assets/Scripts/SampleSceneHandMenu.cs
marlenaklein-msft: I didn't see a way to add a version defines for the sample scripts and this seemed to be the next best thing... Adding the static class to core should resolve this. I'm curious is there way to add version defines for the samples?

marlenaklein-msft:
Closing in favor of new PR: https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/515

UnityProjects/MRTKDevTemplate/Assets/Scripts/SampleSceneHandMenu.cs
marlenaklein-msft: I've made this change and created another PR https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/515

Issue #492, opened 2023-09-27
[FEATURE REQUEST] ObjectManipulator observes XRSocketInteractor selection as expected.


whebertML:
When being selected by an XRSocketInteractor, an ObjectManipulator does not move into the socket transform as expected, like an XRGrabInteractable would.  This is due to the default manipulation logic for ObjectManipulator, which is to retain the relative pose offset of an interactor's attach transform.  So when selected by an XRSocketInteractor, the ObjectManipulator stays where it is instead of moving into the socket.

With the intent to increase MRTK3's compatibility with continuing XRI features, the request is to modify ObjectManipulator's behavior so that it properly observes the XRSocketInteraction selection and moves into the socket transform.

Video demonstrating both current and new behavior with the changes from #567 :

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/58860572/e79cb0fb-60bb-4549-b457-725df993720e

shaynie:
@whebertML please provide a brief description to pass on to Unity for interface suggestions

whebertML:
@AMollis @shaynie 

XRSocketInteractor has direct dependencies on, and calls into, XRGrabInteractable to perform some of the socket features, like snap hovering.  Below are links to several places in the code for reference (links from a mirror of Unity UPM packages for convenience).  Would be preferable if such features were done via some IXRSocketInteractable interface or similar.

https://github.com/needle-mirror/com.unity.xr.interaction.toolkit/blob/master/Runtime/Interaction/Interactors/XRSocketInteractor.cs#L458-L459
https://github.com/needle-mirror/com.unity.xr.interaction.toolkit/blob/master/Runtime/Interaction/Interactors/XRSocketInteractor.cs#L483-L484
https://github.com/needle-mirror/com.unity.xr.interaction.toolkit/blob/master/Runtime/Interaction/Interactors/XRSocketInteractor.cs#L771-L817

shaynie:
@whebertML  please update regarding xri 2.5 vs 2.3

whebertML:
The above noted issues are only within XRI versions post 2.3.2 - I happened to have 2.5 at the time I was investigating.  This does not impact us currently, please disregard.  Though if Unity will be refactoring this, that will be good for future cases potentially.

shaynie:
@whebertML please add unit tests for this

PR #510, opened 2023-10-11
Adding manipulation specializations to handle certain scenarios.

whebertML:
Adds a modular manipulation specialization system to allow for specialized manipulation behavior for ObjectManipulator under certain conditions, like when being selected by a Socket Interactor, that may call for manipulation behavior that is distinct from default.

Fixes: #492

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ManipulationSpecializations/BaseManipulationSpecialization.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ManipulationSpecializations/BaseManipulationSpecialization.cs
@@ -1,0 +1,48 @@
+// Copyright (c) Mixed Reality Toolkit Contributors
+// Licensed under the BSD 3-Clause
+
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.XR.Interaction.Toolkit;
+
+namespace MixedReality.Toolkit.SpatialManipulation
+{
+    /// <summary>
+    /// Abstract base class from which all manipulation specialization behaviors derive.
+    /// </summary>
+    /// <seealso cref="IManipulationSpecialization"/>
+    public abstract class BaseManipulationSpecialization : MonoBehaviour, IManipulationSpecialization
+    {
+        /// <inheritdoc />
+        public abstract bool CanProcessSelection(List<IXRSelectInteractor> interactors,
+                                                 IXRSelectInteractable interactable);
+
+        /// <inheritdoc />
+        public virtual void OnSelectManipulationStarted(List<IXRSelectInteractor> interactors,
+                                                        IXRSelectInteractable interactable,
+                                                        Transform objectTransform,
+                                                        Rigidbody rigidBody)
+        { }
+
+        /// <inheritdoc />
+        public void OnSelectionChanged(List<IXRSelectInteractor> interactors,
+                                       IXRSelectInteractable interactable,
+                                       Transform objectTransform,
+                                       Rigidbody rigidBody)
+        { }
+
+        /// <inheritdoc />
+        public virtual void OnSelectManipulationEnded(List<IXRSelectInteractor> interactors,
+                                                      IXRSelectInteractable interactable,
+                                                      Transform objectTransform,
+                                                      Rigidbody rigidBody)
+        { }
+
+        /// <inheritdoc />
+        public abstract void Process(XRInteractionUpdateOrder.UpdatePhase updatePhase,
+                                     List<IXRSelectInteractor> interactors,
+                                     IXRSelectInteractable interactable,
+                                     Transform objectTransform,
+                                     Rigidbody rigidBody);
+    }
+}
--- a/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ManipulationSpecializations/IManipulationSpecialization.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ManipulationSpecializations/IManipulationSpecialization.cs
@@ -1,0 +1,121 @@
+// Copyright (c) Mixed Reality Toolkit Contributors
+// Licensed under the BSD 3-Clause
+
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.XR.Interaction.Toolkit;
+
+namespace MixedReality.Toolkit.SpatialManipulation
+{
+    /// <summary>
+    /// An interface that allows for specialized manipulation of an <see cref="ObjectManipulator"/>
+    /// under certain circumstances, such as when being selected by an <see cref="XRSocketInteractor"/>,
+    /// that may call for manipulation behavior that is distinct from the default.
+    /// </summary>
+    /// <remarks>
+    /// When active, the manipulation specialization is responsible for complete
+    /// manipulation details, including final position, rotation, and scale of the object
+    /// along with any smoothing and handling of the <see cref="Rigidbody"/> when present.
+    /// A <see cref="BaseManipulationSpecialization"/> is provided for convenience, but implementers
+    /// can use this interface directly as well.
+    /// </remarks>
+    /// <seealso cref="ObjectManipulator"/>
+    public interface IManipulationSpecialization
+    {
+        /// <summary>
+        /// Whether this manipulation specialization is valid for and can process the
+        /// current state of the object with selecting interactors.
+        /// </summary>
+        /// <param name="interactors">
+        /// List of all <see cref="IXRSelectInteractor"/> objects selecting this object.
+        ///</param>
+        /// <param name= "interactable">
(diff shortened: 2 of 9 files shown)


whebertML:
This doesn't yet add any unit tests, but wanted to discuss the viability of the feature and implementation first, then will be adding them as needed.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -609,6 +634,49 @@ private void InstantiateManipulationLogic()
...
+            // Ensure the default set of manipulation specializations is present.
+            gameObject.EnsureComponent<SocketManipulationSpecialization>();
+        }
+
+        private void UpdateSelectManipulationSpecialization()
+        {
+            var previousSpecialization = activeSpecialization;
+            activeSpecialization = null;
+
+            foreach (var specialization in GetComponents<IManipulationSpecialization>())
```
AMollis, line 653: Do we really need multiple specializations?   Seems like your scenario only needs one.

Let's have the developer explicitly set the specialization logic, instead of requiring querying GetComponents.  This query will impact performance even for non-specialized case.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ManipulationSpecializations/BaseManipulationSpecialization.cs
```diff
@@ -0,0 +1,48 @@
...
+using UnityEngine;
+using UnityEngine.XR.Interaction.Toolkit;
+
+namespace MixedReality.Toolkit.SpatialManipulation
+{
+    /// <summary>
+    /// Abstract base class from which all manipulation specialization behaviors derive.
+    /// </summary>
+    /// <seealso cref="IManipulationSpecialization"/>
+    public abstract class BaseManipulationSpecialization : MonoBehaviour, IManipulationSpecialization
```
keveleigh, line 14: Does this need to be a `MonoBehaviour`? I don't see any Unity method callbacks in the base or the implementation.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ManipulationSpecializations/IManipulationSpecialization.cs
```diff
@@ -0,0 +1,121 @@
...
+        ///</param>
+        /// <param name= "interactable">
+        /// The <see cref="IXRSelectInteractable"/> that is being manipulated.
+        /// <remarks>
+        /// The list of attached specialization components will be queried in order during
+        /// <see cref="ObjectManipulator.OnSelectEntered(SelectEnterEventArgs)"/> and the first to
+        /// respond with <c>true</c> will be used as the active manipulation specialization.
+        /// Otherwise, if no specializations respond, the default object manipulation will be used.
+        /// </remarks>
+        public bool CanProcessSelection(List<IXRSelectInteractor> interactors,
```
AMollis, line 40: Can we simplify and reduce the number of parameters? It seems like you only need two parameters `interactors` and `objectmanipulator` and the host transform and rigidBody can be gotten from the object manipulator.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -531,6 +544,10 @@ protected struct LogicImplementation
 
         private bool wasKinematic = false;
 
+        private IManipulationSpecialization activeSpecialization = null;
+
+        private bool specializedManipulationActive => activeSpecialization != null;
```
keveleigh, line 549: nit: PascalCase on properties

I realize this link is for MRTK2, but we weren't able to get around to publishing the coding guidelines for MRTK3 before...you know... 🙃 I'm hoping we'll adopt something similar here though, as these have worked for this project in the past: https://learn.microsoft.com/windows/mixed-reality/mrtk-unity/mrtk2/contributing/coding-guidelines?view=mrtkunity-2022-05#naming-conventions

AMollis (changes requested):
I don't feel comfortable with this "specialization" implementation, as there already exists a specialization mechanism. Can we somehow extend `ManipulationLogic`?

Adding this additional way to specialize an `ObjectManipulator` seems to confuse things. Can we change `ManipulationLogic` to address whatever you need? Or at least follow the pattern set by `ManipulationLogic` and use a "type based" specialization  (see `public struct LogicType`). That way the custom logic lives in a non-MonoBehaviour, and there is no additional performance overhead.

Also have you considered just making a new base class to handle Socket Manipulation?

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -586,6 +603,14 @@ protected override void Awake()
             InstantiateManipulationLogic();
         }
 
+        /// <summary>
+        /// A Unity event function that is called on the frame when a script is enabled just before any of the update methods are called the first time.
+        /// </summary> 
+        private void Start()
```
keveleigh, line 609: Another coding guideline we hadn't set in stone, but one that can cause an easy pitfall for derived classes: the recommendation that Unity messages should be `protected` 
https://github.com/microsoft/Microsoft.Unity.Analyzers/blob/main/doc/UNT0021.md

We didn't do a great job at this in MRTK2, but it had been a point of emphasis while we were developing MRTK3 with an eye for extensibility.

Some additional context at https://github.com/microsoft/MixedRealityToolkit-Unity/pull/11643#discussion_r1240435925

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ManipulationSpecializations/BaseManipulationSpecialization.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        /// <inheritdoc />
+        public abstract bool CanProcessSelection(List<IXRSelectInteractor> interactors,
```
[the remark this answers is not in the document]
whebertML: Will do, didn't notice a lot of examples of wrapping parameters like this, so wasn't sure of convention.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -457,6 +457,19 @@ public LogicType ManipulationLogicTypes
             }
         }
 
+        [SerializeField]
+        [Tooltip("Whether to add the default manipulation specializations to the object if not already present.")]
+        private bool addDefaultManipulationSpecializations = true;
```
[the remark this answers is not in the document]
whebertML: That works too, could leave it up to the dev to add them manually if needed.  Perhaps add a sample demonstrating the use of any specializations (assuming this approach is valid at all - needs discussion).

keveleigh (commented):
First pass

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -738,64 +828,72 @@ public override void ProcessInteractable(XRInteractionUpdateOrder.UpdatePhase up
...
+                {
+                    activeSpecialization.Process(updatePhase, interactorsSelecting, this, HostTransform, rigidBody);
+                }
+                // Default Manipulation Behavior
+                else
                 {
-                    RotateAnchorType rotateType = CurrentInteractionType == InteractionFlags.Near ? RotationAnchorNear : RotationAnchorFar;
-                    bool useCenteredAnchor = rotateType == RotateAnchorType.RotateAboutObjectCenter;
-                    bool isOneHanded = interactorsSelecting.Count == 1;
+                    // Evaluate user input in the UI Update() function.
```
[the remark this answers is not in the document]
whebertML: This else was to fall back into the original, "default" manipulation logic, so to contain all the original default logic and not tie it into the encapsulating if/else

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ManipulationSpecializations/BaseManipulationSpecialization.cs
whebertML: As discussed in meeting, I only decided to try use MonoBehaviour due to it easily providing user tweakable fields - like the Constraint system.  But may go back to just having this be a "type selected" system (non-MonoBehaviour) like ManipulationLogic.  Or see if there would be a less invasive way to provide for this type of thing all together (from discussion).

whebertML:
Closing this PR in favor of an alternative, less invasive, approach in #526

PR #515, opened 2023-10-18
Add Precompile Statements to Prevent Warnings in Unity 2023 (Refactored)

marlenaklein-msft:
Following @keveleigh's suggestion to refactor this PR https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/506.

FIxes:

- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/502

### DIFF
--- a/org.mixedrealitytoolkit.core/Utilities/FindObjectUtility.cs
+++ b/org.mixedrealitytoolkit.core/Utilities/FindObjectUtility.cs
@@ -1,0 +1,47 @@
+// Copyright (c) Mixed Reality Toolkit Contributors
+// Licensed under the BSD 3-Clause
+
+using UnityEngine;
+using System;
+
+namespace MixedReality.Toolkit
+{
+    public static class FindObjectUtility
+    {
+        public static T FindFirstObjectByType<T>(bool includeInactive = false) where T : Component
+        {
+#if UNITY_2021_3_18_OR_NEWER
+        return UnityEngine.Object.FindFirstObjectByType<T>(includeInactive ? FindObjectsInactive.Include : FindObjectsInactive.Exclude);
+#else
+        return UnityEngine.Object.FindObjectOfType<T>(includeInactive);
+#endif
+        }
+
+        public static T FindAnyObjectByType<T>(bool includeInactive = false) where T : Component
+        {
+#if UNITY_2021_3_18_OR_NEWER
+        return UnityEngine.Object.FindAnyObjectByType<T>(includeInactive ? FindObjectsInactive.Include : FindObjectsInactive.Exclude);
+#else
+        return UnityEngine.Object.FindObjectOfType<T>(includeInactive);
+#endif
+        }
+
+        public static T[] FindObjectsOfType<T>(bool includeInactive = false, bool sort = true) where T : Component
+        {
+#if UNITY_2021_3_18_OR_NEWER
+        return UnityEngine.Object.FindObjectsByType<T>(includeInactive ? FindObjectsInactive.Include : FindObjectsInactive.Exclude, sort ? FindObjectsSortMode.InstanceID : FindObjectsSortMode.None);
+#else
+            return UnityEngine.Object.FindObjectsOfType<T>(includeInactive);
+#endif
+        }
+
+        public static UnityEngine.Object[] FindObjectsOfType(Type type, bool includeInactive = false, bool sort = true)
+        {
+#if UNITY_2021_3_18_OR_NEWER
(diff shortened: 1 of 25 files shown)


org.mixedrealitytoolkit.core/Utilities/FindObjectUtility.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace MixedReality.Toolkit
+{
+    /// <summary>
+    /// A static utility used to avoid deprecated Find Object functions in favor of replacements introduced in Unity >= 2021.3.18. 
+    /// </summary>
+    public static class FindObjectUtility
+    {
+
+        /// <summary>
```
AMollis, line 15: Comment Style.

Can you the unity version information in a `<remarks>` tag.  And in the `<summary>` tag don't call out the implementation details, instead describe what the functions do in more general terms.

AMollis (approved):

org.mixedrealitytoolkit.core/Utilities/FindObjectUtility.cs
marlenaklein-msft: will do

AMollis (approved):

org.mixedrealitytoolkit.uxcore/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "org.mixedrealitytoolkit.uxcore",
-  "version": "3.1.0-development",
+  "version": "3.1.1-development",
```
keveleigh, line 3: I think the package dependencies need to be updated throughout as well, since this depends on a new API in the Core package.

org.mixedrealitytoolkit.input/InteractionModes/InteractionModeManager.cs
```diff
@@ -71,7 +71,7 @@ public static InteractionModeManager Instance
         public void InitializeControllers()
         {
             controllerMapping.Clear();
-            foreach (XRController xrController in FindObjectsOfType<XRController>())
+            foreach (XRController xrController in FindObjectUtility.FindObjectsOfType<XRController>())
```
keveleigh, line 74: Looks like the Input package.json wasn't updated

org.mixedrealitytoolkit.core/Utilities/FindObjectUtility.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// Returns all objects matching the specified type.
+        /// </summary>
+        /// <remarks>
+        /// If Unity >= 2021.3.18, calls FindObjectsByType. Otherwise calls FindObjectsOfType.
+        /// </remarks>
+        /// <param name="includeInactive">If true, inactive objects will be included in the search. False by default.</param>
+        /// <param name="sort">If false, results will not sorted by InstanceID. True by default.</param>
+        public static T[] FindObjectsOfType<T>(bool includeInactive = false, bool sort = true) where T : Component
```
keveleigh, line 55: nit: do we want to align the naming? it looks like the single object methods use "_**By**_ Type" and the multi-object methods use "_**Of**_ Type".

I don't feel strongly one way or the other, but it seems like Unity is moving to use "By" for all their new methods, so it could be good to standardize in that direction too

keveleigh (changes requested):

org.mixedrealitytoolkit.uxcore/package.json
marlenaklein-msft: good point!

AMollis: Good call!

org.mixedrealitytoolkit.input/InteractionModes/InteractionModeManager.cs
marlenaklein-msft: I think it was already bumped and hopefully this should go in before the next release

AMollis: We haven't shipped and are not shipping Input 3.0.1 yet.   The package.json is already 3.0.1, last shipped version was 3.0.0

org.mixedrealitytoolkit.core/Utilities/FindObjectUtility.cs
marlenaklein-msft: will change to make it consistent :)

org.mixedrealitytoolkit.input/InteractionModes/InteractionModeManager.cs
keveleigh: Oops! Missed that

keveleigh (approved):

PR #526, opened 2023-10-20
Adding Rigidbody MovementType to ObjectManipulator

whebertML:
This adds XRI's XRBaseInteractable.MovementType to ObjectManipulator so devs can specify the movement type of the rigid body during manipulation.

Currently the behavior always forces non-kinematic "Velocity Tracking" when a rigid body is attached, setting velocity and angular velocity to move the rigid body.  This would allow the dev to specify other types of movement, like kinematic, as is provided in XRI.  This also observes the XRBaseInteractor.selectedInteractableMovementTypeOverride when an interactor overrides the movement type.

Resolves:  #532 

Related to being more compatible with XRI #67

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
@@ -162,6 +162,36 @@ namespace MixedReality.Toolkit.SpatialManipulation
         }
 
         [SerializeField]
+        [Tooltip("The preferred movement type of the rigid body during manipulation. Selecting interactors may override this and define a different movement type.")]
+        private MovementType rigidbodyMovementType = MovementType.VelocityTracking;
+
+        /// <summary>
+        /// The <see cref="XRBaseInteractable.MovementType"/> of the rigid body during manipulation.
+        /// </summary>
+        /// <remarks>
+        /// This is the default <see cref="XRBaseInteractable.MovementType"/> of the rigid body during manipulation,
+        /// however, the type can be overridden by an interactor with <see cref="XRBaseInteractor.selectedInteractableMovementTypeOverride"/>.
+        /// The overriding value, if any, will be set by the most recent selecting interactor that has an override.
+        /// </remarks>
+        public MovementType RigidbodyMovementType
+        {
+            get => rigidbodyMovementType;
+            set
+            {
+                if (rigidbodyMovementType != value)
+                {
+                    rigidbodyMovementType = value;
+
+                    if (isSelected && rigidBody != null)
+                    {
+                        UpdateCurrentRigidbodyMovementType();
+                        SetupRigidbodyManipulation();
+                    }
+                }
+            }
+        }
+
+        [SerializeField]
         [Tooltip(
             "Apply torque to control orientation of the body")]
         private bool applyTorque = true;
@@ -517,6 +547,16 @@ namespace MixedReality.Toolkit.SpatialManipulation
         /// </summary>
         protected LogicImplementation ManipulationLogic { get; private set; }
 
+        /// <summary>
+        /// The current <see cref="XRBaseInteractable.MovementType"/> of the rigid body during manipulation.
+        /// </summary>
+        /// <remarks>
+        /// This defaults to <see cref="RigidbodyMovementType"/>, but can be overridden
+        /// by an interactor with <see cref="XRBaseInteractor.selectedInteractableMovementTypeOverride"/>.
+        /// The overriding value, if any, will be set by the most recent selecting interactor that has an override.
+        /// </remarks>
+        protected MovementType CurrentRigidbodyMovementType { get; private set; } = MovementType.VelocityTracking;
+
         #endregion Protected Properties
 
         #region Private Properties
@@ -524,6 +564,8 @@ namespace MixedReality.Toolkit.SpatialManipulation
         private ITransformSmoothingLogic smoothingLogic;
 
         private bool ShouldSmooth => (IsGrabSelected && SmoothingNear) || (!IsGrabSelected && SmoothingFar);
+
+        private bool UseForces => rigidBody != null && !rigidBody.isKinematic;
 
         private Rigidbody rigidBody;
 
@@ -647,7 +689,6 @@ namespace MixedReality.Toolkit.SpatialManipulation
         private Vector3 referenceFrameLastPos;
 
         private MixedRealityTransform targetTransform;
-        private bool useForces;
 
         private static readonly ProfilerMarker OnSelectEnteredPerfMarker =
             new ProfilerMarker("[MRTK] ObjectManipulator.OnSelectEntered");
@@ -669,16 +710,19 @@ namespace MixedReality.Toolkit.SpatialManipulation
             {
                 base.OnSelectEntered(args);
 
-                // Only record rigidbody settings if this is the *first*
-                // selection event! Otherwise, we'll record the during-interaction
-                // rigidbody information, which we've already dirtied.
-                if (rigidBody != null && interactorsSelecting.Count == 1)
+                if (rigidBody != null)
                 {
-                    wasGravity = rigidBody.useGravity;
-                    wasKinematic = rigidBody.isKinematic;
+                    // Only record rigidbody settings if this is the *first*
+                    // selection event! Otherwise, we'll record the during-interaction
+                    // rigidbody information, which we've already dirtied.
+                    if (interactorsSelecting.Count == 1)
+                    {
+                        wasGravity = rigidBody.useGravity;
+                        wasKinematic = rigidBody.isKinematic;
+                    }
 
-                    rigidBody.useGravity = false;
-                    rigidBody.isKinematic = false;
+                    UpdateCurrentRigidbodyMovementType();
+                    SetupRigidbodyManipulation();
                 }
 
                 targetTransform = new MixedRealityTransform(HostTransform.position, HostTransform.rotation, HostTransform.localScale);
@@ -691,8 +735,6 @@ namespace MixedReality.Toolkit.SpatialManipulation
                 {
                     constraintsManager.OnManipulationStarted(targetTransform);
                 }
-
-                useForces = rigidBody != null && !rigidBody.isKinematic;
 
                 // ideally, the reference frame should be that of the camera. Here the interactorObject transform is the best available alternative.
                 referenceFrameTransform = GetReferenceFrameTransform(args);
@@ -712,9 +754,17 @@ namespace MixedReality.Toolkit.SpatialManipulation
 
                 // Only release the rigidbody (restore rigidbody settings/configuration)
                 // if this is the last select event!
-                if (rigidBody != null && interactorsSelecting.Count == 0)
+                if (rigidBody != null)
                 {
-                    ReleaseRigidBody(rigidBody.velocity, rigidBody.angularVelocity);
+                    if (interactorsSelecting.Count == 0)
+                    {
+                        ReleaseRigidBody(rigidBody.velocity, rigidBody.angularVelocity);
+                    }
+                    else
+                    {
+                        UpdateCurrentRigidbodyMovementType();
+                        SetupRigidbodyManipulation();
+                    }
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -811,15 +849,15 @@ private void ApplyTargetTransform()
...
                 HostTransform.localScale = targetTransform.Scale;
             }
             else
             {
                 // There is a Rigidbody. Potential different paths for near vs far manipulation
-                if (!useForces)
+                if (!UseForces)
                 {
                     rigidBody.MovePosition(targetTransform.Position);
                     rigidBody.MoveRotation(targetTransform.Rotation);
```
whebertML, line 863: Note:  I left these original rigiBody.MovePosition/Rotation here in ApplyTargetTransform() which operates in update loop, though technically I think you're supposed to do these in the Fixed update loop.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -161,6 +161,32 @@ public InteractionFlags AllowedInteractionTypes
...
+        public MovementType RigidbodyMovementType
+        {
+            get => rigidbodyMovementType;
+            set
+            {
+                rigidbodyMovementType = value;
+
+                if (isSelected && rigidBody != null)
+                {
+                    UpdateRigidbodyMovementType();
```
AMollis, line 184: Should we check if the value has actually changed here, to avoid unnecessary work?

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -525,12 +551,16 @@ protected struct LogicImplementation
...
+        private bool UseForces => rigidBody != null && !rigidBody.isKinematic;
+
         private Rigidbody rigidBody;
 
         private bool wasGravity = false;
 
         private bool wasKinematic = false;
 
+        private MovementType currentRigidbodyMovementType;
+
```
AMollis, line 563: Can we explicitly set a default value for `currentRigidbodyMovementType`?

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -161,6 +161,32 @@ public InteractionFlags AllowedInteractionTypes
             set => allowedInteractionTypes = value;
         }
 
+        [SerializeField]
+        [Tooltip("The movement type of the rigid body during manipulation.")]
+        private MovementType rigidbodyMovementType = MovementType.VelocityTracking;
```
AMollis, line 166: Since this value can be overridden by an interactor, I think we should rename `rigidbodyMovementType` so to indicate that this value might not used.  

Perhaps `defaultRigidbodyMovementType`  or `preferredRigidbodyMovementType`

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -525,12 +551,16 @@ protected struct LogicImplementation
...
 
+        private bool UseForces => rigidBody != null && !rigidBody.isKinematic;
+
         private Rigidbody rigidBody;
 
         private bool wasGravity = false;
 
         private bool wasKinematic = false;
 
+        private MovementType currentRigidbodyMovementType;
```
AMollis, line 562: I wonder if we should expose this value to subclasses?  Make a property instead?

```cs
/// <summary>
/// The selected <see cref="XRBaseInteractable.MovementType"/>of the rigid body during manipulation.
/// </summary>
/// <remarks>
/// This defaults to <see cref="DefaultRigidbodyMovementType"/>, but can be override 
/// by an interactor with <see cref="XRBaseInteractor.selectedInteractableMovementTypeOverride"/>.
/// </remarks>
protected MovementType CurrentRigidbodyMovementType { get; private set; } = // some default value
```

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -939,11 +977,38 @@ protected virtual void ModifyTargetPose(ref MixedRealityTransform targetPose, re
...
+            targetPose.Position = (applySmoothing && !UseForces) ? smoothingLogic.SmoothPosition(HostTransform.position, targetPose.Position, moveLerpTime, Time.deltaTime) : targetPose.Position;
+            targetPose.Rotation = (applySmoothing && !UseForces) ? smoothingLogic.SmoothRotation(HostTransform.rotation, targetPose.Rotation, rotateLerpTime, Time.deltaTime) : targetPose.Rotation;
             targetPose.Scale = applySmoothing ? smoothingLogic.SmoothScale(HostTransform.localScale, targetPose.Scale, scaleLerpTime, Time.deltaTime) : targetPose.Scale;
         }
 
+        private void UpdateRigidbodyMovementType()
+        {
+            // Find the most recent interactor that has a movement type override, if any, using reverse order of selecting interactors.
+            // This matches XRI's treatment of overrides as the default behavior.
+            MovementType? movementTypeOverride = null;
```
AMollis, line 989: We can avoid the overhead of boxing a value type...  I believe you can refactor to this...

```cs
            currentRigidbodyMovementType = rigidbodyMovementType;
            for (var index = interactorsSelecting.Count - 1; index >= 0; --index)
            {
                var xrBaseInteractor = interactorsSelecting[index] as XRBaseInteractor;
                if (xrBaseInteractor != null && xrBaseInteractor.selectedInteractableMovementTypeOverride.HasValue)
                {
                    currentRigidbodyMovementType = xrBaseInteractor.selectedInteractableMovementTypeOverride.Value;
                    break;
                }
            }
```

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -939,11 +977,38 @@ protected virtual void ModifyTargetPose(ref MixedRealityTransform targetPose, re
...
-            targetPose.Position = (applySmoothing && !useForces) ? smoothingLogic.SmoothPosition(HostTransform.position, targetPose.Position, moveLerpTime, Time.deltaTime) : targetPose.Position;
-            targetPose.Rotation = (applySmoothing && !useForces) ? smoothingLogic.SmoothRotation(HostTransform.rotation, targetPose.Rotation, rotateLerpTime, Time.deltaTime) : targetPose.Rotation;
+            targetPose.Position = (applySmoothing && !UseForces) ? smoothingLogic.SmoothPosition(HostTransform.position, targetPose.Position, moveLerpTime, Time.deltaTime) : targetPose.Position;
+            targetPose.Rotation = (applySmoothing && !UseForces) ? smoothingLogic.SmoothRotation(HostTransform.rotation, targetPose.Rotation, rotateLerpTime, Time.deltaTime) : targetPose.Rotation;
             targetPose.Scale = applySmoothing ? smoothingLogic.SmoothScale(HostTransform.localScale, targetPose.Scale, scaleLerpTime, Time.deltaTime) : targetPose.Scale;
         }
 
+        private void UpdateRigidbodyMovementType()
+        {
+            // Find the most recent interactor that has a movement type override, if any, using reverse order of selecting interactors.
```
AMollis, line 987: Just a question.... does the reverse order mean that the override is controlled by the last selecting interactor?

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
```diff
@@ -939,11 +977,38 @@ protected virtual void ModifyTargetPose(ref MixedRealityTransform targetPose, re
...
             bool applySmoothing = ShouldSmooth && smoothingLogic != null;
 
-            targetPose.Position = (applySmoothing && !useForces) ? smoothingLogic.SmoothPosition(HostTransform.position, targetPose.Position, moveLerpTime, Time.deltaTime) : targetPose.Position;
-            targetPose.Rotation = (applySmoothing && !useForces) ? smoothingLogic.SmoothRotation(HostTransform.rotation, targetPose.Rotation, rotateLerpTime, Time.deltaTime) : targetPose.Rotation;
+            targetPose.Position = (applySmoothing && !UseForces) ? smoothingLogic.SmoothPosition(HostTransform.position, targetPose.Position, moveLerpTime, Time.deltaTime) : targetPose.Position;
+            targetPose.Rotation = (applySmoothing && !UseForces) ? smoothingLogic.SmoothRotation(HostTransform.rotation, targetPose.Rotation, rotateLerpTime, Time.deltaTime) : targetPose.Rotation;
             targetPose.Scale = applySmoothing ? smoothingLogic.SmoothScale(HostTransform.localScale, targetPose.Scale, scaleLerpTime, Time.deltaTime) : targetPose.Scale;
         }
 
+        private void UpdateRigidbodyMovementType()
```
AMollis, line 985: Can we rename to match the variable name that is changing?  Rename `UpdateRigidbodyMovementType` to `UpdateCurrentRigidbodyMovementType`

AMollis (changes requested):
I like the change.    Just some minor requests

whebertML:
Thanks for the review @AMollis!  Have a couple of fires internally, but will update and respond soon as I can.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
whebertML: Good point!  Done.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
whebertML: Good point, done.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
whebertML: Done

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
whebertML: I like that idea, useful to subclasses.  Done.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
whebertML: Yes.  And that matches XRI's XRGrabInteractable's selection of an override.

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
whebertML: Done

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
whebertML: My personal take is that isn't necessary in order to be clear.  For example, the XRI XRGrabInteractable's property is just named "movementType", though it is understood to be potentially overridden by an interactor without expressing that in the name.

In our case, at least, I had added a description of the overriding behavior in the <remarks> section of the documentation on the RigidbodyMovementType property.  Would that be sufficient?

AMollis: Ok :)  

Can you all add something to the tooltip?  So developers in the inspector have some helpful documentations as well.

example:

```cs
[Tooltip("The preferred movement type of the rigid body during manipulation. Selecting interactors may override this and define a different movement type.")]
```

AMollis (changes requested):
Thanks for the updates.  Small tooltip request

org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/ObjectManipulator.cs
whebertML: Done.  Also, since this involves new API, updated the minor version of the spatial manipulation package.

AMollis (approved):

PR #535, opened 2023-11-01
Update serialized applicationIdentifiers

keveleigh:
These match the Project Name and Organization Name elsewhere in the serialization. They only update on build, I believe, so it just wasn't updated when the previous updates were made.

### DIFF
--- a/UnityProjects/MRTKDevTemplate/ProjectSettings/ProjectSettings.asset
+++ b/UnityProjects/MRTKDevTemplate/ProjectSettings/ProjectSettings.asset
@@ -156,13 +156,13 @@ PlayerSettings:
   androidSupportedAspectRatio: 1
   androidMaxAspectRatio: 2.1
   applicationIdentifier:
-    Android: com.MixedRealityToolkitOrganization.MRTK3Sample
-    Standalone: com.Mixed-Reality-Toolkit-Organization.MRTK3-Sample
+    Android: org.MixedRealityToolkit.MRTK3Sample
+    Standalone: org.MixedRealityToolkit.MRTK3Sample
   buildNumber:
     Standalone: 0
     iPhone: 0
     tvOS: 0
-  overrideDefaultApplicationIdentifier: 0
+  overrideDefaultApplicationIdentifier: 1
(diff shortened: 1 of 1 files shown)


UnityProjects/MRTKDevTemplate/ProjectSettings/ProjectSettings.asset
```diff
@@ -156,8 +156,8 @@ PlayerSettings:
   androidSupportedAspectRatio: 1
   androidMaxAspectRatio: 2.1
   applicationIdentifier:
-    Android: com.Microsoft.MRTK3Sample
-    Standalone: com.Microsoft.MRTK3Sample
+    Android: com.MixedRealityToolkitOrganization.MRTK3Sample
```
AMollis, line 159: should be org.MixedRealityToolkit  :)  For the registered MixedRealityToolkit.org domain.

UnityProjects/MRTKDevTemplate/ProjectSettings/ProjectSettings.asset
```diff
@@ -156,8 +156,8 @@ PlayerSettings:
...
+    Standalone: com.Mixed-Reality-Toolkit-Organization.MRTK3-Sample
```
AMollis, line 160: org.MixedRealityToolkit

AMollis (changes requested):
I think technically we should be referencing domain registrations that we own.   So replace the com with org, and remove "Organization"

MixedRealityToolkit.org

UnityProjects/MRTKDevTemplate/ProjectSettings/ProjectSettings.asset
keveleigh: Overrode the auto-genned strings to match our domain

UnityProjects/MRTKDevTemplate/ProjectSettings/ProjectSettings.asset
keveleigh: Overrode the auto-genned strings to match our domain

AMollis (approved):

Issue #540, opened 2023-11-06
[FEATURE REQUEST] Allow VirtualizedScrollRectList to be reused


ahmed-shariff:
## Problem

Based on my understanding, during runtime, once a VirtualizedScrollRectList is created, it cannot be reset. I would like to be able to reuse the menu for different list of items.

The OnValidate method already seems to do this. A solution could be to make this public method
https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/f669f5aef4141ed3f949a2913a40e386c5444b2e/org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs#L314-L324

AMollis:
Hi @ahmed-shariff  thank you for your request.   This seems like a valid feature worth investigating.

PR #541, opened 2023-11-06
Allow resetting VirtualizedDynamicScrollRectList (#540)

ahmed-shariff:
In response to #540 

This PR add a `Reset` public method to `VirtualizedScrollRectList`. The `OnValidate` method already contains the resetting functionality, this PR simply moves that to a new `Reset` method accessible from outside the class.


org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
```diff
@@ -531,6 +524,22 @@ public void SetItemCount(int newCount)
             visibleValid = false;
             UpdateScrollView(scroll);
         }
+
+        /// <summary>
+        /// Resets the VirtualizedScrollRectList. This may remove items aready
+        /// visible and create new items.
+        /// </summary>
+        public void Reset()
```
AMollis, line 532: Reset() is a MonoBehaviour editor mode event function.   Can you please rename this metho to avoid confusion.   Perhaps `ResetLayout`?

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
```diff
@@ -394,7 +387,7 @@ private void Initialize()
 
         private void InitializePool()
         {
-            // Support resetting everything from OnValidate
+            // Support resetting everything from Reset
```
AMollis, line 390: See other comment....please update this comment once the Reset funciton is renamed

AMollis (changes requested):
Please add a unit test that validates the "reset" functionality.

ahmed-shariff:
Thank you for reviewing the PR. I could use some insight into how the tests are structured. The `VirtualizedScrollRectList` has a "test" in the dev template: https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/f669f5aef4141ed3f949a2913a40e386c5444b2e/UnityProjects/MRTKDevTemplate/Assets/Scripts/VirtualizedScrollRectListTester.cs#L14

This has some of the functionality I would use in the unit test. Would it be ok to move that to the package? or should I leave that script as it is?

AMollis:
> Thank you for reviewing the PR. I could use some insight into how the tests are structured. The `VirtualizedScrollRectList` has a "test" in the dev template:
> 
> https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/f669f5aef4141ed3f949a2913a40e386c5444b2e/UnityProjects/MRTKDevTemplate/Assets/Scripts/VirtualizedScrollRectListTester.cs#L14
> 
> This has some of the functionality I would use in the unit test. Would it be ok to move that to the package? or should I leave that script as it is?

@ahmed-shariff , it looks like the original owner did not write Unit Tests for `VirtualizedDynamicScrollRectList`, which is unfortunate.  Given this, I'll be okay with one of the follow:

1. Setup a new unit test class here https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/tree/main/org.mixedrealitytoolkit.uxcore/Tests/Runtime and write a test to validate the reset functionality.
2. Update the Sample scene, [VirtualizedScrollRectList](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/main/UnityProjects/MRTKDevTemplate/Assets/Scenes/Experimental/VirtualizedScrollRectList.unity), to demonstrate the "resetting" functionality.  Maybe add a button to resize and reset the list? 

I worry since this is not common/obvious functionality, there's a higher risk for regression if we don't provide a method to validate.

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
ahmed-shariff: It is now called reset layout

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
```diff
@@ -236,18 +254,38 @@ public float Scroll
...
+        /// invoked. When the value is set the list of items get reset
+        /// <see cref="ResetLayout"/>.
         /// </summary>
-        public Action<GameObject, int> OnInvisible { get; set; }
+        public Action<GameObject, int> OnInvisible
+        {
+            get => onInvisible;
+            set
+            {
+                onInvisible = value;
```
AMollis, line 284: I think we should check for a value change here.

```cs
    if (onInvisible != value)
    {
        onInvisible = value;
        // If this is changeing, it means how the items are populated is changing, hence reset.
        ResetLayout();
    }
```

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
```diff
@@ -236,18 +254,38 @@ public float Scroll
...
+        /// visible. When the value is set the list of items get reset.
+        /// <see cref="ResetLayout"/>.
         /// </summary>
-        public Action<GameObject, int> OnVisible { get; set; }
+        public Action<GameObject, int> OnVisible
+        {
+            get => onVisible;
+            set
+            {
+                onVisible = value;
```
AMollis, line 265: check for value change

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
```diff
@@ -190,6 +193,21 @@ public float Scroll
...
+        /// The pool of list item objects will be composed of this Prefab. When
+        /// the value changes the items are all repopulated.
+        /// <see cref="ResetLayout"/>
+        /// </summary>
+        public GameObject Prefab
+        {
+            get => prefab;
+            set
+            {
+                prefab = value;
```
AMollis, line 206: check for value change

AMollis (changes requested):
Thank you so much for writing those Unit Test :)   Thank you !

I have some very minor suggestions....

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
ahmed-shariff: That's a good catch. Made the updates & added that to the tests as well.

AMollis (dismissed):

AMollis:
@shaynie , @marlenaklein-msft , @keveleigh .  

Once the AZP build completes, I'll merge unless you have feedback

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
```diff
@@ -531,6 +580,27 @@ public void SetItemCount(int newCount)
...
+        {
+            // We only want to reset things if it has already been initialized,
+            // we don't want to initialize it prematurely!
+            if (initialized == false)
+            {
+                return;
+            }
+
+            // We don't want to reset multiple times in one frame.
+            if (resetCalled)
```
keveleigh, line 598: I don't see where `resetCalled` is set to `true`. Should it be below here, right before the `StartCoroutine`?

keveleigh (changes requested):

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
ahmed-shariff: Thank you for catching that. :D

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
```diff
@@ -484,6 +524,15 @@ private void MakeVisible(int i)
             OnVisible?.Invoke(go, i);
         }
 
+        private IEnumerator ResetLayoutAtEndOfFrame()
+        {
+            yield return new WaitForEndOfFrame();
```
keveleigh, line 529: I think this _may_ cause an issue with the batch mode test runner on our PR validation 😞 

https://docs.unity3d.com/ScriptReference/WaitForEndOfFrame.html

> Note: This coroutine is not invoked on the editor in batch mode.

Seems like there's a suggestion on to use `yield return Application.isBatchMode ? null : WaitForEndOfFrame();` at https://forum.unity.com/threads/do-not-use-waitforendofframe.883648/#post-5807410

keveleigh: I'd also be curious to know how this works in-editor when play mode isn't running, since this code path is also triggered by `OnValidate()`

marlenaklein-msft (dismissed):

AMollis (changes requested):
Changing my vote.... the Unit Tests seem to be failing.  Please fix Unit Tests

```
MixedReality.Toolkit.UX.Runtime.Tests.VirtualizedScrollRectListTests.TestVirtualizedScrollRectList_ResetLayout
  Non of the expected items were found in the scollable list (set1).
  Expected: True
  But was:  False
```

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
ahmed-shariff: I did not know about that. Thank you for pointing that out, pushing with that change. I was building on the code that was already there. In retrospect, `yield return null` would have the same outcome.

As for the OnValidate code path, from what I understand, it is what allows the UI to change when a value on the inspector is updated. There's probably a more optimal way to handle that without completely resetting it, but I don't want to touch that right now 😅

ahmed-shariff:
> Changing my vote.... the Unit Tests seem to be failing. Please fix Unit Tests
> 
> ```
> MixedReality.Toolkit.UX.Runtime.Tests.VirtualizedScrollRectListTests.TestVirtualizedScrollRectList_ResetLayout
>   Non of the expected items were found in the scollable list (set1).
>   Expected: True
>   But was:  False
> ```

Not sure what is causing that. The tests were passing on my end. I have made a few updates to when frames are skipped, can we test if they pass again?

AMollis:
> > Changing my vote.... the Unit Tests seem to be failing. Please fix Unit Tests
> > ```
> > MixedReality.Toolkit.UX.Runtime.Tests.VirtualizedScrollRectListTests.TestVirtualizedScrollRectList_ResetLayout
> >   Non of the expected items were found in the scollable list (set1).
> >   Expected: True
> >   But was:  False
> > ```
> 
> Not sure what is causing that. The tests were passing on my end. I have made a few updates to when frames are skipped, can we test if they pass again?

I'm guessing the Unit Tests were failing because of the previous `WaitForEndOfFrame()` calls which aren't supported in batch mode.  The CI pipeline is running the Unit Tests in batch mode.

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
AMollis: @keveleigh are you ok with this change?

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
```diff
@@ -531,6 +580,28 @@ public void SetItemCount(int newCount)
...
+            {
+                return;
+            }
+
+            // We don't want to reset multiple times in one frame.
+            if (resetCalled)
+            {
+                return;
+            }
+            resetCalled = true;
```
AMollis, line 602: nit. add a new line after the `{`

AMollis (approved):
I approve.   Once @keveleigh approve we will merge

org.mixedrealitytoolkit.uxcore/Experimental/List/VirtualizedScrollRectList.cs
```diff
@@ -236,18 +257,44 @@ public float Scroll
...
-        public Action<GameObject, int> OnInvisible { get; set; }
+        public Action<GameObject, int> OnInvisible
+        {
+            get => onInvisible;
+            set
+            {
+                if (onInvisible != value)
+                {
+                    onInvisible = value;
+                    // If this is changeing, it means how the items are populated is changing, hence reset.
```
keveleigh, line 293:
```suggestion
                    // If this is changing, it means how the items are populated is changing, hence reset.
```

keveleigh (approved):

keveleigh (approved):

AMollis (approved):

AMollis (approved):

Issue #544, opened 2023-11-08
[BUG] The cubes in some scenes in the MRTKDevTemplate project change color after being grabbed.


whebertML:
## Describe the bug

Minor cosmetic issue - some of the colored cubes in the HandInteractionExamples and DisabledInteractorsExample scene change color during, and after, manipulation while others retain color (as intended?).

## To reproduce

1. Go to the HandInteractionExamples scene in the MRTKDevTemplate project
2. Run the scene in Editor
3. Grab, manipulate, and release the colored cubes on the right.
4. Notice the lime and purple colored cubes change primary color once grabbed and reset to different colors on release.

## Expected behavior

All cubes retain their color

### Current behavior
Lime cube idle
![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/58860572/33f93e28-9ac7-496b-af5a-f60357972e72)
Lime cube grabbed (changes to emerald grabbed material)
![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/58860572/33646aee-4ff6-430b-874f-0320476db6af)
Lime cube released (changed to emerald idle material)
![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/58860572/0915df6d-c340-48b1-8a6f-c4e255b5a30c)

### After fix
Lime cube grabbed
![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/58860572/cc9bbcfc-0395-4acc-ac02-3a6e951cd1c5)
Lime cube released
![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/58860572/e64932bc-03ea-4c86-95fa-bf3de833a3d0)

PR #545, opened 2023-11-08
Fix minor issue of cubes changing color in some scenes when grabbed

whebertML:
Fixes: #544 

Fixes a minor cosmetic / quality issue with some of the MRTKDevTemplate samples where some of the colored cubes change color once grabbed and released.  The intended behavior would be for the cubes to retain their color to maintain the uniqueness of each cube during and after manipulation, even after falling off the edge and being restored.

Just a minor issue with a simple material reference fix.

### DIFF
--- a/org.mixedrealitytoolkit.extendedassets/Materials/ObjectManipulationGrabbedLime.mat
+++ b/org.mixedrealitytoolkit.extendedassets/Materials/ObjectManipulationGrabbedLime.mat
@@ -1,0 +1,228 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!21 &2100000
+Material:
+  serializedVersion: 8
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_Name: ObjectManipulationGrabbedLime
+  m_Shader: {fileID: 4800000, guid: c331f6c43a2ef0945864cb668f2653c9, type: 3}
+  m_ValidKeywords:
+  - _DISABLE_ALBEDO_MAP
+  - _INNER_GLOW
+  - _PROXIMITY_LIGHT
+  - _USE_WORLD_SCALE
+  m_InvalidKeywords: []
+  m_LightmapFlags: 4
+  m_EnableInstancingVariants: 1
+  m_DoubleSidedGI: 0
+  m_CustomRenderQueue: 2000
+  stringTagMap:
+    DisableBatching: False
+    RenderType: Opaque
+  disabledShaderPasses: []
+  m_SavedProperties:
+    serializedVersion: 3
+    m_TexEnvs:
+    - _BumpMap:
+        m_Texture: {fileID: 0}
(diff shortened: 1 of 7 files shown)


AMollis (changes requested):
@whebertML can you please add screen shots of the before and after changes

Also, please add more details to the PR description.  In particular I'm curious about the need for this change.

org.mixedrealitytoolkit.extendedassets/Materials/ObjectManipulationGrabbedLime.mat
AMollis, line 1: If this is for samples only, please move these to the Sample project.

keveleigh: There are already two "Grabbed" materials in Extended Assets, and I'm not sure I agree they should be moved. This is basically the self-described purpose of this package: "This package offers additional assets helpful for example scenes"

AMollis: Makes sense thanks for setting my straight :)

whebertML: Done.

I originally made the new materials to sit alongside the other "grabbed" materials for some of those cubes that lives in extended assets.  Happy to just create new materials in the project itself without need to modify the package.  👍

AMollis: Sorry @whebertML .... I didn't realize that there were already Object Manipulator materials in here.   @keveleigh  set me straight.   Apologizes, you can undo your last commit, since there are already sample materials in the package.. I'm sorry for not realizing this. 

FWIW...why are sample materials in a UPM package?  What was the thinking here?

whebertML:
Before and after screen shots placed in issue description #544

org.mixedrealitytoolkit.extendedassets/Materials/ObjectManipulationGrabbedLime.mat
keveleigh: > FWIW...why are sample materials in a UPM package? What was the thinking here?

100% open to revisiting this decision, but the idea was predicated around two things: sample distribution and prototyping enablement.

Under the assumption that we'd distribute our samples the same way we did in MRTK2 (as embedded samples in UPM packages, but trying to stay away from the idea of a samples-specific package, so we'd instead be embedding samples in their contextual packages), we realized we needed _some_ shared set of assets to reduce duplication. The approach of an "extended assets" package allowed us to avoid bloating those packages due to embedded sample assets and instead prompt the user to install this "extended assets" package when they imported the samples.

Secondarily, we thought it could be useful to build up a set of semi-polished visual components that could be used in prototyping scenarios, without bloating the "standard assets" package and without requiring the prototyper to import all the samples into their local project.

That's my recollection at least 😅

whebertML: @AMollis @keveleigh   Materials moved back to extendedassets

AMollis (approved):

Issue #523, opened 2023-10-19
[BUG] Slider handle not correcly set when value set by script


anonymous2585:
## Describe the bug

I have two sliders that misbehave in my project.

Sliders goes from 0 to 3 with a default value of 1 and are desactivated by default in the scene. I have a call to `slider.Value = 2.5f` during the same frame as the `gameObject.SetActive(true)`. Sliders are children of multiples canvas layouts.
The fill background is correctly set, but not the handle;

![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/148471165/0746015e-f0d3-4969-8f11-c733a2432885)

I added some logs, the issue comes from the Slider `TrackDirection` that is no corrrect when my script call `slider.Value = x`.

![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/148471165/46be27b4-1e4d-45fb-923a-514c4a6a1655)
Logs 1 to 3 are from the MRTK initializing (multiple times) the slider.
The 4th logs comes from my script triggering `slider.Value = 2.5f`, handle is incorrectly placed.
The 5th logs come from a mouse click on the slider at the fill end position. The handle is then correcly placed.

As you can see, `TrackDirection` is incorrect during the 4 first logs.

The problem comes from this line:
https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/e30b350c131501084a4f9003cba1361248717f66/org.mixedrealitytoolkit.uxcore/Slider/Visuals/CanvasSliderVisuals.cs#L202C18-L202C18

Is it possible to do the computation "in canvas space" with anchors (like the `fillVisual`) instead of setting the absolute world position?

## Expected behavior

The handle is set at the same place than the fill ends.

## Your setup (please complete the following information)

- Unity Version 2023.1.16f1
- MRTK UX Components 3.1.0

anonymous2585:
@marlenaklein-msft
If you replace the `UpdateHandle` function with this, it solves the issue for me.

```c#
        // Update the things that depend on the slider value.
        void UpdateHandle(float value)
        {
            switch (SliderDirection)
            {
                case Direction.LeftToRight:
                    fillVisual.anchorMax = new Vector2(value, 1.0f);
                    handle.anchorMin = new Vector2(value, handle.anchorMin.y);
                    handle.anchorMax = new Vector2(value, handle.anchorMax.y);
                    break;
[...]

PR #549, opened 2023-11-10
Fix Canvas Slider Handle Positioning and Impose MinMax Limits to Slider Value

marlenaklein-msft:
Implement fixes related to setting the slider value programmatically: canvas slider handle calculations now use canvas positioning, aligning with fillVisual. Issue and workaround reported by @anonymous2585, https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/523.

Also adds a check to limit Slider value within the range of min and max values, removes arbitrary prefab value.

### DIFF
--- a/org.mixedrealitytoolkit.uxcore/Slider/Slider.cs
+++ b/org.mixedrealitytoolkit.uxcore/Slider/Slider.cs
@@ -155,7 +155,7 @@ namespace MixedReality.Toolkit.UX
                 {
                     if (value < minValue || value > maxValue)
                     {
-                        Debug.Log($"{this.value} is outside of the slider range.");
+                        Debug.Log($"{value} is outside of the slider range.");
                     }
                     else
                     {
--- a/org.mixedrealitytoolkit.uxcore/Slider/Slider.cs
+++ b/org.mixedrealitytoolkit.uxcore/Slider/Slider.cs
@@ -151,7 +151,11 @@ namespace MixedReality.Toolkit.UX
             get => value;
             set
             {
-                if (this.value != value)
+                if (value < minValue || value > maxValue)
+                {
+                    Debug.Log($"{this.value} is outside of the slider range.");
+                }
+                else
                 {
                     var oldSliderValue = this.value;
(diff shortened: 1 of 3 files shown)


ms-RistoRK:
LGTM, ship it!

org.mixedrealitytoolkit.uxcore/Slider/Slider.cs
```diff
@@ -153,9 +153,16 @@ public float Value
             {
                 if (this.value != value)
                 {
-                    var oldSliderValue = this.value;
-                    this.value = value;
-                    OnValueUpdated.Invoke(new SliderEventData(oldSliderValue, value));
+                    if (value < minValue || value > maxValue)
+                    {
+                        Debug.Log($"{value} is outside of the slider range.");
```
AMollis, line 158: I think we should instead clamp the value. 

`value = Mathf.Clamp(value, minValue, maxValue)`

Also, I'm not sure how I feel about this log message, maybe just remove?

AMollis (changes requested):

org.mixedrealitytoolkit.uxcore/Slider/Slider.cs
marlenaklein-msft: will change

AMollis (approved):

AMollis (changes requested):
Since you're change uxcomponents...please update its patch number to 3.0.1

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/main/com.microsoft.mrtk.uxcomponents/package.json

AMollis:
> Since you're change uxcomponents...please update its patch number to 3.0.1
> 
> https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/main/com.microsoft.mrtk.uxcomponents/package.json

Ugh...I was looking at the wrong package.json

AMollis (approved):

anonymous2585:
Awesome, thank you for this PR! 🎉

Issue #543, opened 2023-11-08
BoundsHandleInteractable throws errors on empty GameObject[BUG]


johannes-weidenfeller:
## Describe the bug

If an empty GameObject with an attached BoundsControl component is instantiated during runtime, the BoundsHandleControl component of the BoundingBoxWithHandles Bounds Visual Prefab throws the following error:

_transform.localScale assign attempt for 'ScaleHandle' is not valid. Input localScale is { Infinity, 100.000000, 100.000000 }._

These errors are not thrown when the object is initially placed in the scene instead of instantiated during runtime.

## To reproduce

1. Create a prefab of an empty GameObject with a BoundsControl component and assign the BoundingBoxWithHandles prefab to the Bounds Visual Prefab of the BoundsControl component
2. Instantiate the prefab during runtime

## Expected behavior

The transforms of the bounds handles should not be set if the bounds are zero which seems to work for scene-placed objects.

## Screenshots

![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/54772951/d28c162d-d83d-415f-a5e1-29a8b03d49c5)

## Your setup (please complete the following information)

- Unity Version 2021.3.19f1
- MRTK Version 3.0.0-pre.18

AMollis:
@johannes-weidenfeller, please update the GA versions for MRTK3, and try again.

AMollis:
@marlenaklein-msft , I think you looked at bounds control before?  Does this seem like something we addressed in MRTK3 GA?

marlenaklein-msft:
The previous issue with infinite bounds was specific to flattened bounds and resolved here https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/12. I am guessing it may be something similar where a vector is not being updated... will look into this

marlenaklein-msft:
Resolved with this PR https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/558

PR #558, opened 2023-11-21
Fix Bugs with Bounding Box Initialized at Runtime

marlenaklein-msft:
Fixes the errors reported in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/543, as well as current behavior where an empty game object initialized at runtime will have unreasonable bounds values for one frame and a non-empty prefab like the coffee cup will have handles incorrectly enabled for one frame:

image

This change ensures that the LateUpdate in BoundsHandleInteractable is not called before Update when bounding boxes are initialized at runtime. This prevents incorrect vector values and division by zero, and the other odd behavior for that first frame.


org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
```diff
@@ -102,6 +104,14 @@ protected override void Awake()
             HideOnStartup();
         }
 
+        /// <summary>
+        /// A Unity event function that is called every frame, if this object is enabled.
+        /// </summary>
+        protected virtual void Update()
+        {
+            hasUpdated = true;
```
AMollis, line 112: Can we validate the initialization state in LateUpdate() without adding an additional flag?

AMollis (changes requested):

marlenaklein-msft:
**Issue 1: Handles Incorrectly Appear** 
The handlesEnabled variable is applied during SqueezeableBoxVisuals' Update. If an object with handles disabled is initialized at runtime, BoundsHandleInteractable's LateUpdate is called first and the handles will incorrectly appear for that first frame.
**Solution:** Call Update from Awake in Squeezable Box Visuals, make a protected flag, or delay LateUpdate until after Update in BoundsHandleInteractable.

**Issue 2: Errors logged for invalid scale**
In BoundsHandleInteractable LateUpdate, the local scale is set to 1/transform.lossyscale. When transform.lossyscale is 0 we get an error from division by 0. The lossyscale being 0 is attributable to the grandparent object (SqueezableBoxVisuals) which has its bounds set by BoundsControl's ComputeBounds which is called from BoundsControl's Update. 
**Solution:** Check explicitly for division by 0, change the call ComputeBounds (complicated to refactor), or delay LateUpdate until after Update in BoundsHandleInteractable.

**Issue 3: Star-shaped explosion when colliders are enabled with invalid scale.** 
**Solution:** Move enabling of colliders to Update or delay LateUpdate until after Update in BoundsHandleInteractable.

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
```diff
@@ -117,27 +134,18 @@ public void Start()
...
-            }
-
             // Maintain the aspect ratio/proportion of the handles, globally.
             transform.localScale = Vector3.one;
-            transform.localScale = new Vector3(1.0f / transform.lossyScale.x,
-                                               1.0f / transform.lossyScale.y,
-                                               1.0f / transform.lossyScale.z);
-            
+            transform.localScale = new Vector3(1.0f / (transform.lossyScale.x == 0 ? 1 : transform.lossyScale.x),
+                                1.0f / (transform.lossyScale.y == 0 ? 1 : transform.lossyScale.y),
```
AMollis, line 140: nit: please indent only four spaces

```cs
            transform.localScale = new Vector3(1.0f / (transform.lossyScale.x == 0 ? 1 : transform.lossyScale.x),
                1.0f / (transform.lossyScale.y == 0 ? 1 : transform.lossyScale.y),
                1.0f / (transform.lossyScale.z == 0 ? 1 : transform.lossyScale.z));
```

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SqueezableBoxVisuals.cs
```diff
@@ -194,6 +194,7 @@ private void Awake()
 
             // Compute flatten vector at startup.
             flattenVector = BoundsCalculator.CalculateFlattenVector(transform.lossyScale);
+            Update();
```
AMollis, line 197: We shouldn't be calling Update() in Awake().   Update() should only be called directly by the Unity.

What were you trying to initialize here?

AMollis (changes requested):

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SqueezableBoxVisuals.cs
AMollis: Maybe instead of Update(), the squeeze handles should be updated in LateUpdate(), to match the bounds control.

Or can we initialize the squeeze box visuals to some initial state here, without executing all of update()...for example, can we do this?


```cs
float initialActiveFocus = 0;
float initialActivation = 0;
float initialShrink = 0;
WritePropertyValues(boundsRenderer, propertyBlock,, initialActiveFocus , initialActivation , initialShrink);
```

AMollis (changes requested):

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
```diff
@@ -102,6 +102,23 @@ protected override void Awake()
             HideOnStartup();
         }
 
+        /// <summary>
+        /// A Unity event function that is called every frame, if this object is enabled.
+        /// </summary>
+        protected virtual void Update()
+        {
+            // Do our IsOccluded "setter" in Update so we don't do this multiple times a frame.
+            if (IsOccluded != wasOccludedLastFrame)
```
AMollis, line 111: The SqueezeBoxVisuals update BoundsHandleInteractable::IsOccluded in Update(). To ensure that BoundsHandleInteractable has a valid IsOccluded value for the frame, BoundsHandleInteractable has to use IsOccluded  in LateUpdate() and not Update()....
 
So we shouldn't move this code.

AMollis: I think if you remove the call to SqueezableBoxVisuals::Update() in  SqueezableBoxVisuals::Awake() .... this code can be in LateUpdate().   That is, if SqueezableBoxVisuals::Update() is not called, then IsOccluded is not updated. Which means `if (IsOccluded != wasOccludedLastFrame)` won't evaluate to `true`.

AMollis (changes requested):

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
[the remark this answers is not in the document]
AMollis: Never mind I see `SqueezeBoxVisuals::OnEnable` is calling this.

However, I think this could be causing part of your bug.  Since `SqueezeBoxVisuals::OnEnable` and  `BoundsHandleInteractable::Awake" calls `HideOnStartup()`, and `HideOnStartup()` sets wasOccludedLastFrame to true. LateUpdate() incorrectly using the IsOccluded value too soon.  

I think we need to re-think `HideOnStartup` that was added with this PR....https://github.com/microsoft/MixedRealityToolkit-Unity/pull/11698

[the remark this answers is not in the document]
AMollis: If we need to keep HideOnStartup, then should set IsOccluded to true in HideOnStartup...that should fix stuff

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
AMollis: Opps... found another issue.  Since `SqueezeBoxVisuals::OnEnable` and  `BoundsHandleInteractable::Awake` calls `HideOnStartup()`, and `HideOnStartup()` sets `wasOccludedLastFrame` to `true`. `LateUpdate()` incorrectly uses the `IsOccluded` value too soon.  

We re-think `HideOnStartup` that was added with this PR....https://github.com/microsoft/MixedRealityToolkit-Unity/pull/11698

If we need to keep HideOnStartup, then should set IsOccluded to true in HideOnStartup...that should fix stuff

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
marlenaklein-msft: done

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SqueezableBoxVisuals.cs
marlenaklein-msft: with force occlusion function, this is no longer needed

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
```diff
@@ -161,6 +163,23 @@ public void HideOnStartup()
             wasOccludedLastFrame = true;
         }
 
+        /// <summary>
+        /// Sets <see cref="IsOccluded"/> to true, and forces handling of occlusion immediately."/>
```
AMollis, line 167: Sorry this is probably my fault... remove this bogus stuff at the end of the sentence...

 "/>

AMollis (approved):
One comment :)

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
```diff
@@ -151,6 +152,7 @@ private float MaxComponent(Vector3 v)
         /// <summary>
         /// Occludes the handle so it is not initially visible when it should start disabled.
         /// </summary>
+        [Obsolete("Force hiding handles is no longer supported. Use IsOccluded instead.")]
```
AMollis, line 155: ack .... please fix my typo :)

"Force hiding is no longer supported. Use IsOccluded instead."

AMollis (approved):

PR #567, opened 2023-11-29
ObjectManipulator's ManipulationLogic observes XRSocketInteractor

### LINKED ISSUE #492: [FEATURE REQUEST] ObjectManipulator observes XRSocketInteractor selection as expected.

whebertML:
* Modifies ManipulationLogic for position and rotation to properly observe XRSocketInteractor.
* Turns off object smoothing when selected by XRSocketInteractor.
* Also observes XRRayInteractor forceGrab property for position, moving the object to the ray interactor's force grab attach transform.

Resolves #492

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/MoveLogics/ManipulationLogic.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/MoveLogics/ManipulationLogic.cs
@@ -23,6 +23,16 @@ namespace MixedReality.Toolkit.SpatialManipulation
         protected int NumInteractors { get; private set; }
 
         /// <summary>
+        /// Whether the object is selected by a singular <see cref="XRSocketInteractor"/>.
+        /// </summary>
+        protected bool SelectedBySocket { get; private set; }
+
+        /// <summary>
+        /// Whether the object is force grabbed by a singular <see cref="XRRayInteractor"/>.
+        /// </summary>
+        protected bool ForceGrabbed { get; private set; }
+
+        /// <summary>
         /// Setup the manipulation logic. Called automatically by Update if the number of interactor points has changed.
         /// </summary>
         /// <param name="interactors">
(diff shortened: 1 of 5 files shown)


david-c-kline (dismissed):
Approved pending unit test success :smile:

AMollis (approved):

whebertML:
Unit tests added.

AMollis (approved):

anonymous2585:
@whebertML thank you for this improvement!

I tested it and it works but the option "Hover Socket Snapping" is not observed by the ObjectManipulator when it is set to `true`: the object don't snap to the socket attach position before user release select.

Can you check if you have the same behaviour and open another issue if confirmed?

whebertML:
@anonymous2585   Thank you for testing the feature!

I believe the MRTK3 project itself is still only dependent on XRI version 2.3.0, which doesn't yet contain the `Hover Socket Snapping` property of `XRSocketInteractor`.  Not sure when that was introduces (XRI 2.5.2?), but such a feature could be added when the project upgrades it's dependency on a newer XRI I would think.

anonymous2585:
@whebertML You are right! I didn't noticed [Hover Socket Snapping was added only in XRI 2.5.0](https://docs.unity3d.com/Packages/com.unity.xr.interaction.toolkit@2.5/changelog/CHANGELOG.html#250---2023-08-17)
> Added snap transformations to XRSocketInteractor on hover.

Yes, we use XRI 2.5.2.

Maybe it's feasible to support it when the installed XRI package is `>= 2.5.0`?

whebertML:
@anonymous2585 Definitely technical possibly of course.  Please feel free to add a feature request and it will be discussed.

Issue #572, opened 2023-12-01
[BUG] Initial scale of `BoundsHandleInteractable` is not taken into account on initialization


anonymous2585:
## Describe the bug

The initial scale relative to the parent is not taken into account on handles initialization.

I also expect the initial transform scale on each handles in the prefab to be taken in account for the scaling. That way, a user can do a prefab with big handles, one other with smaller handles...

I also think it's better to take the minimum (actual is maximum) of the parent lossy scale, so handles doesn't risk to collide on the shorter side.

## To reproduce

1. Import [this Unity package](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/files/13527314/BoundsHandlesSize.zip)
2. Open scene `BoundsControlIssue`
3. Open prefab `BoundingBoxWithTraditionalHandlesNoMaintainGlobalSize`, see the scale of handles is very big (100) and maintain global scale is unckecked
4. Close prefab
5. Press play
6. See issue (handles scale should have been as big as in the prefab, in fact too big to be usable)
7. Open prefab `BoundingBoxWithTraditionalHandlesNoMaintainGlobalSize`, set the scale of handles to a standard value (10)
8. Close prefab
9. Press play
10. See realtime scale is still the same as in step 6 and still too small

## Expected behavior

The initial handles scale is the same as if a user grabed the prefab (scaled 1x1x1, handles scale to 10 like in step 10) and extended it to a 15x6x10 scale:
![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/148471165/e9a23d85-8144-4b9b-bf68-348e3fd593b3)

## Screenshots

Actual result is a big box with really small handles hard to grab:
![image](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/148471165/ef8f2f49-90a5-4c99-93ac-54f809b9e8b1)

## Setup

- Unity Version 2021.3.21f1
- MRTK Version or Commit: main branch [a09f5cd](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/a09f5cd98b2b47b668bcc6fc775bcc7fd3735eb7)

## Target platform

- HoloLens 2
- HP Reverb G2
- iOS AR
- OpenXR

marlenaklein-msft:
[...]

PR #598, opened 2024-01-02
Allow Variable Bounding Box Handle Sizes

marlenaklein-msft:
Currently, all BoundsHandleInteractables are set to a global scale of 1x1x1 at runtime (and may then scale with the bounding box if MaintainGlobalSize is unselected). The initial scale of BoundsHandleInteractable is never taken into account when scaling the handles. This change sets the global scale to the max(x,y,z) of the initial BoundsHandleInteractable scale, instead of 1, so a user may now change the size of the handles in the prefab in order to make bounding boxes with larger or smaller handles. PR also makes changes to the BoundsControlRuntimeExample scene to give an example of this.

image

[3x3x3 Handles Scaled Prefab from Sample Scene]

Addresses https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/572

Spatial Manipulation package version number has already been bumped since last release, so no update is needed.

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
@@ -91,7 +91,7 @@ namespace MixedReality.Toolkit.SpatialManipulation
 
         private float initialParentScale;
 
-        private float initialHandleScale;
+        private float initialLocalScale;
 
         /// <inheritdoc/>
         protected override void Awake()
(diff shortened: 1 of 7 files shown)


org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
```diff
@@ -91,6 +91,8 @@ public BoundsControl BoundsControlRoot
 
         private float initialParentScale;
 
+        private float initialHandleScale;
```
AMollis, line 94: Can we rename this to initialLocalScale

AMollis (changes requested):
One minor name change request.  Also can you please attach a screen shot of the new prefab you've added?

Thank you

AMollis:
@marlenaklein-msft , did you also check if the version numbers need updating?   Can you please add a comment to the PR description stating you've verified the package versions are correct?  Thank you

AMollis (approved):

PR #608, opened 2024-01-09
Fix warning and event triggered on disabled StatefulInteractable after changing speech settings (#591)

### LINKED ISSUE #591: [BUG] Disabled `StatefulInteractable` are interactable after set `SpeechRecognitionKeyword` or `AllowSelectByVoice`
## Describe the bug

When setting the properties `SpeechRecognitionKeyword` or `AllowSelectByVoice` of a disabled `StatefulInteractable`, a warning is triggered and the interactable starts again to trigger clicks events even though it's disabled.

This is due to the button unregister and register back to the interaction manager. 

## To reproduce

1. Open the CanvasExample scene
2. Choose one of the "Action Button" and add this script:
```C#
using MixedReality.Toolkit;
using System;
using System.Collections;
using System.Collections.Generic;
using UnityEngine;

[RequireComponent(typeof(StatefulInteractable))]
public class VoiceRuntimeChange : MonoBehaviour
{
    private IEnumerator Start()
    {
        StatefulInteractable interactable = GetComponent<StatefulInteractable>();
        interactable.OnClicked.AddListener(OnClicked);
        interactable.enabled = false;
        yield return new WaitForSeconds(1.0f);
        interactable.SpeechRecognitionKeyword = "Hello world"; // trigger a warning and the button is enabled
        yield return new WaitForSeconds(1.0f);
        interactable.AllowSelectByVoice = false; // the button is enabled
    }

    private void OnClicked()
    {
        Debug.Log("Button clicked");
    }
}
```
4. Press Play
5. Wait 1 second and see the warning
6. Click on the button and see the debug log triggered

## Expected behavior

The interactable is doesn't trigger the warning and doesn't trigger click events

## Your setup

- Unity Version: 2021.3.21f1
- MRTK Version or Commit: current main branch [ec5c141](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/ec5c14129925745a2ad167efb9eecae988b06020)

## Target platform

- HoloLens 2
- OpenXR

## Additional context

The fix is to check
```C#
if (Application.isPlaying && isActiveAndEnabled)
```
before unregister and register the interactable. This will also fix editor script using these properties.
The interactable will register himself on enable (in unity XRI code).

Side note: why is `VoiceRequiresFocus` private set? I needed to set it from code but I couldn't.

anonymous2585:
Fixes #591 

+ Fix the warning and wrong event triggered
+ Add non-regression tests
+ Make VoiceRequiresFocus public

### DIFF
--- a/org.mixedrealitytoolkit.core/Interactables/StatefulInteractable.cs
+++ b/org.mixedrealitytoolkit.core/Interactables/StatefulInteractable.cs
@@ -111,9 +111,15 @@ namespace MixedReality.Toolkit
                 if (value != allowSelectByVoice)
                 {
                     // Unregister and re-register the interactable to update the speech interactor with latest info
-                    interactionManager.UnregisterInteractable(this as IXRInteractable);
+                    if (Application.isPlaying && isActiveAndEnabled)
+                    {
+                        interactionManager.UnregisterInteractable(this as IXRInteractable);
+                    }
                     allowSelectByVoice = value;
-                    interactionManager.RegisterInteractable(this as IXRInteractable);
+                    if (Application.isPlaying && isActiveAndEnabled)
+                    {
+                        interactionManager.RegisterInteractable(this as IXRInteractable);
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


shaynie (approved):
Thanks for the fix!

PR #609, opened 2024-01-09
Make GazePinchInteractor.DependentInteractor a XRBaseInteractor (Fix #605)

### LINKED ISSUE #605: [BUG] Deprecated `XRBaseControllerInteractor` in XRI 3.0.0-pre.1 breaks `GazePinchInteractor`
## Describe the bug

When upgrading to XRI 3.0.0-pre.1 I get the following error:

```
Library\PackageCache\org.mixedrealitytoolkit.input@719fc273754c\Interactors\GazePinch\GazePinchInteractor.cs(350,45): error CS8121: An expression of type 'XRBaseControllerInteractor' cannot be handled by a pattern of type 'FuzzyGazeInteractor'.
```

- Unity Version 2022.3.9f
- MRTK INput Version: 3.0.0

## Additional context
In XRI 3.0.0 `XRBaseControllerInteractor` is being deprecated and replaced with `XRBaseInputInteractor` (https://docs.unity3d.com/Packages/com.unity.xr.interaction.toolkit@3.0/changelog/CHANGELOG.html).

## Potential fix
Looking through the code, `GazePinchInteractor.DependentInteractor` can be a `XRBaseInteractor` as only the `IXRInteractor.GetValidTargets` is being used in `GazePinchInteractor`. Both `XRBaseControllerInteractor` and `XRBaseInputInteractor` extend this as well. This also should not conflict with `GazeInteractor` as it extends `XRRayInteractor`, which is extending the appropriate class on both version (`XRBaseControllerInteractor` in 2.5.2 and `XRBaseInputInteractor` in 3.0.0-pre.1)

ahmed-shariff:
Fixes #605
XRI 3 deprecates `XRBaseControllerInteractor` in favor `XRBaseInputInteractor`. This breaks `GazePointInteractor` as  `GazePinchInteractor.DependentInteractor` is a `XRBaseControllerInteractor`. This PR makes this variable an `XRBaseInteractor`. This shouldn't cause any conflicts as only its `IXRInteractor` interface implementations are used.

### DIFF
--- a/org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
+++ b/org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
@@ -70,12 +70,12 @@ namespace MixedReality.Toolkit.Input
 
         [SerializeField]
         [Tooltip("The interactor we're using to query potential gaze pinch targets")]
-        private XRBaseControllerInteractor dependentInteractor;
+        private XRBaseInteractor dependentInteractor;
 
         /// <summary>
         /// The pose source representing the ray this interactor uses for aiming and positioning.
         /// </summary>
-        protected XRBaseControllerInteractor DependentInteractor { get => dependentInteractor; set => dependentInteractor = value; }
+        protected XRBaseInteractor DependentInteractor { get => dependentInteractor; set => dependentInteractor = value; }
 
         [SerializeField]
         [Range(0, 1)]


org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -70,12 +70,12 @@ public class GazePinchInteractor :
 
         [SerializeField]
         [Tooltip("The interactor we're using to query potential gaze pinch targets")]
-        private XRBaseControllerInteractor dependentInteractor;
+        private XRBaseInteractor dependentInteractor;
```
AMollis, line 73: This looks like a breaking change.   We probably want to hold off on this change until the next major release.

AMollis (changes requested):

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ahmed-shariff: As mentioned above only the `IXRInteracator` implementations (namely `GetValidTargets` and the casting of extended interfaces) are used. But if someone is extending this class and depending on implementations specific to `XRBaseControllerInteractor`, then, yes, this change could cause issues. Otherwise, functionally, this change would not make any difference; it works on both XRI 2.5.2 and XRI 3.0.0-pre.1.

shaynie:
There is a feature branch with the update to XRI 3.0 that will address this.  see #645

PR #614, opened 2024-01-11
Fix empty SpeechRecognitionKeyword breaking all speech keyword system (#612)

### LINKED ISSUE #612: [BUG] Put an empty/whitespace string in `SpeechRecognitionKeyword` break the whole speech system
## Describe the bug

If a user or a script set the `SpeechRecognitionKeyword` on a `StatefulInteractable` to an empty/whitespace string (and AllowSelectByVoice is still true), the whole speech system stops recognizint keywords, without any warning or error.

If you have a script (in my case a `LocalizedStringEvent` for Unity Localization package) that briefly set an empty string and then set the correct string, the bug still remains and speech won't work at all.

## To reproduce

1. Open the CanvasExemple scene
2. Choose one button and set the SpeechRecognitionKeyword to an empty field
3. Build & Run on Hololens
4. Look at another button, say "select" and see nothing happen

## Expected behavior

If one button SpeechRecognitionKeyword is set to an empty string, other buttons still work.

## Your setup

- Unity Version: 2021.3.21f1
- MRTK Version or Commit: current main branch [3707957](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/3707957ef03b4287bd76967705af02cec6c1fb8a)

## Target platform

- HoloLens 2

## Additional context

The issue is that the `SpeechInteractor` allows to register interactable with an empty string put in the `keywordDictionary` and then the Unity `KeywordRecognizer` doesn't work.

anonymous2585:
Fix #612 

If user or code set the SpeechRecognitionKeyword to an empty or whitespace string, the SpeechInteractor won't register the blank keyword in the keywordDictionary and won't transmit it to the subsystem.

### DIFF
--- a/org.mixedrealitytoolkit.input/Interactors/Speech/SpeechInteractor.cs
+++ b/org.mixedrealitytoolkit.input/Interactors/Speech/SpeechInteractor.cs
@@ -58,23 +58,26 @@ namespace MixedReality.Toolkit.Input
         /// <param name="keyword">The keyword to listen for.</param>
         public void RegisterInteractable(StatefulInteractable interactable, string keyword)
         {
-            keyword = keyword.ToLower();
-            if (keywordDictionary.TryGetValue(keyword, out List<StatefulInteractable> interactableList))
+            if (!string.IsNullOrWhiteSpace(keyword))
             {
-                interactableList.Add(interactable);
-            }
-            else
-            {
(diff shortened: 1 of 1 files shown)


marlenaklein-msft (approved):
Thanks for the fix!

Issue #574, opened 2023-12-01
[BUG] Hand ray and reticule are hidden when scrolling a scrollable from a disabled child interactable


anonymous2585:
## Describe the bug

Hand ray and reticule are hidden when scrolling a scrollable from a disabled child interactable.

## To reproduce

1. Open the scene `Experimental/ScrollingExample`
2. Unfold all children of the GameObject `VerticalAndHortizontalScrolling`
3. Uncheck `Is Interactable` on some of the children called `Action Button (x)`
4. Press play and test with the hand simulator, see ray hidden when the scrolling starts with the ray on one of the disabled buttons (edit: now I have no scroll at all, but the ray doesn't disappear)
4. Build the and run on Hololens 2, see same thing

## Expected behavior

Hand ray and reticule are always visible when scrolling in a scrollable.

## Setup

- Unity Version 2021.3.21f1
- MRTK Version or Commit: main branch [a09f5cd](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/a09f5cd98b2b47b668bcc6fc775bcc7fd3735eb7)

## Target platform (please complete the following information)

- HoloLens 2
- HP Reverb G2
- OpenXR

AMollis:
@anonymous2585  can you provide a video of the failure?

anonymous2585:
Here is an Hololens video:
https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/148471165/482c7c52-e613-4204-8b0d-fa0c555d86b5
At first, I clicked on a not interactable (IsInteractable = false) button, ray desappeared but movements are ok. Then I show it's ok with an activated button.

Here is a view in editor:
![BugDisabledButtonsEditor](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/148471165/de5bf848-d341-4594-826a-ac597ea0fe38)
At first I clicked on a not interactable (IsInteractable = false) button, movements didn't work at all. Then I show it's ok with an activated button.

AMollis:
Hi @anonymous2585,

Thank you for reporting this issue or requesting this feature. We appreciate your feedback and interest in MRTK3.

We have reviewed your issue or feature request and confirmed that it is valid. We have added it to our backlog and will prioritize it accordingly.

[...]

PR #626, opened 2024-01-29
Auto disable button's colliders when `StatefulInteractable` is disabled

AMollis:
# Overview

This pull request aims to make it easier to disable an interactable’s colliders when `StatefiulInteractable.IsInteractable` is false or when the `StatefiulInteractable.enabled` is false. This is because list scrolling would break if the user attempted to scroll the list while focusing the disabled interactable. This is because the interactable’s collider was still enabled and blocking ray casts to the list’s backplate.

To work around this, this change adds a new button component, `StatefulInteractableColliderToggle`, and new property, `RectTransformColliderFitter.CanToggleCollider`. The new `RectTransformColliderFitter` property, `CanToggleCollider`, will allow code to turn off the fitter’s automatic enabling/disabling when the collider's visibility changes. The new behavior class, `StatefulInteractableColliderToggle`, automatically disables or enables an interactable’s colliders based on this interactor’s enablement state. This component, to prevent visibility changes causing false collider enables, also sets the `RectTransformColliderFitter.CanToggleCollider` value.

Unit tests were added to validate this new functionality.

# Fixes
- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/574

# Alternative Considerations

1. Adding auto-collider-disablement logic to the interactable itself was considered. However, in order to prevent the RectTransformColliderFitter from changing the collider’s enabled state as well, the interactable would introduce a reference to RectTransformColliderFitter. This would add additional complication to the interactable code that most scenarios don’t require. As such, it was opted to move logic to an isolated component that could be selectively added to the appropriate scenarios.

2. Adding auto-collider-disable logic to `RectTransformColliderFitter`. This would fit nicely into `RectTransformColliderFitter`, since `RectTransformColliderFitter` is already controlling collider's enablement.  However, for this to work correctly, `RectTransformColliderFitter` would need a reference to the interactable so it could query the interactor's enabled state. Leaking interactor knowledge into `RectTransformColliderFitter` was not desired, as `RectTransformColliderFitter`'s main purpose is to resize a collider to fitter a transform.  This resize logic can be useful for non-interactable components.  So leaking interactable knowledge into `RectTransformColliderFitter` would muddy its original intent, to resize colliders.

### DIFF
--- a/org.mixedrealitytoolkit.uxcore/Button/StatefulInteractableColliderToggle.cs
+++ b/org.mixedrealitytoolkit.uxcore/Button/StatefulInteractableColliderToggle.cs
@@ -1,0 +1,106 @@
+// Copyright (c) Mixed Reality Toolkit Contributors
+// Licensed under the BSD 3-Clause
+
+using Unity.Profiling;
+using UnityEngine;
+using UnityEngine.EventSystems;
+using UnityEngine.Serialization;
+using UnityEngine.UI;
+
+namespace MixedReality.Toolkit.UX
+{
+    /// <summary>
+    /// The 'InteractableColliderToggle' class is responsible for managing the state of the collider 
+    /// associated with a StatefulInteractable in the Unity UI. It enables or disables the collider based on 
+    /// the state of the associated 'StatefulInteractable' object.
+    /// </summary>
+    [AddComponentMenu("MRTK/UX/Stateful Interactable Collider Toggle")]
+    public class StatefulInteractableColliderToggle : UIBehaviour
+    {       
+        [SerializeField]
+        [Tooltip("The StatefulInteractable to enable or disable the collider based on the Interactable's state.")]
+        private StatefulInteractable statefulInteractable;
+
+        /// <summary>
+        /// The StatefulInteractable to enable or disable the collider based on the Interactable's state.
+        /// </summary>
+        public StatefulInteractable StatefulInteractable
+        {
+            get => statefulInteractable;
+            set => statefulInteractable = value;
+        }
+        
+        [SerializeField]
+        [Tooltip("The collider to enable or disable based on the Interactable's state.")]
+        private BoxCollider thisCollider;
+
+        /// <summary>
+        /// The collider to enable or disable based on the Interactable's state.
+        /// </summary>
+        public BoxCollider ThisCollider
+        {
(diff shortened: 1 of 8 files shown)


marlenaklein-msft (approved):

AMollis:
@keveleigh , @whebertML , @ghazen-ml , @shaynie , @david-c-kline ... can I get a review please.  Please provide feedback by Friday.  :)

shaynie (approved):
Dependency changes look good.
Unit tests run clean.
Approved

org.mixedrealitytoolkit.uxcore/Button/StatefulInteractableColliderToggle.cs
```diff
@@ -0,0 +1,113 @@
...
+        [Tooltip("The StatefulInteractable to enable or disable the collider based on the interactable's enabled state.")]
+        private StatefulInteractable statefulInteractable;
+
+        /// <summary>
+        /// The <see cref="StatefulInteractable"/> to enable or disable the collider based on the interactable's enabled state.
+        /// </summary>
+        public StatefulInteractable StatefulInteractable
+        {
+            get => statefulInteractable;
+            set => statefulInteractable = value;
```
whebertML, line 29: @AMollis Wouldn't we want to run UpdateCollider(); when the StatefulInteractable is set as well, like we do when setting ColliderFitter?  It could already be disabled, say, and we'd need to set colliders disabled at that moment?

whebertML (commented):
One question, otherwise looks good.

org.mixedrealitytoolkit.uxcore/Button/StatefulInteractableColliderToggle.cs
whebertML: Also I guess also subscribing to the newly set StateInteractable's OnEnable/OnDisable since we are supporting having it set via code as well as in Inspector?

AMollis: Oh good point @whebertML .   I'll make that change

AMollis:
@whebertML , thank you for the review.  I've made changes to the PR

whebertML (approved):

PR #635, opened 2024-02-08
Adding SimpleButton + demo scene

ms-RistoRK:
The SimpleButton structure is inspired in Mesh's MasterButtonRoot which is a simplified version of MRTK3's Action Button.  Note: Materials (including shaders) are those existing in MRTK3 or its associated packages.


org.mixedrealitytoolkit.uxcomponents/Editor/CreateElementMenus.cs
```diff
@@ -210,6 +214,19 @@ private static void CreateActionButton(MenuCommand menuCommand)
             CreateElementFromPath(ActionButtonPath, menuCommand);
         }
 
+        [MenuItem("GameObject/UI/MRTK/Simple Button")]
+        private static void CreateSimpleButton(MenuCommand menuCommand)
+        {
+            GameObject simpleButton = CreateElementFromPath(SimpleButtonPath, menuCommand);
+            GameObject simpleButtonContentGameObject = simpleButton.transform.GetChild(3).gameObject;
```
AMollis, line 221: Please add the `Undo.RecordObject` operations, so callers can undo actions on components.

Follow pattern of `CreateActionButtonWide`

AMollis (changes requested):
Can you please share some screen shots of the new scenes?

org.mixedrealitytoolkit.uxcomponents/Editor/CreateElementMenus.cs
ms-RistoRK: Added them in latest commit.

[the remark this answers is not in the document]
ms-RistoRK: Yes, I do it (in latest commit) with:
```
            foreach (Transform child in simpleButton.transform)
            {
                if (child.name.Equals("Content"))
                {
                    simpleButtonContentGameObject = child.gameObject;
                }
            }
```
**Note 1**:  The reason I didn't use `gameObject.GetComponentsInChildren<TextMeshProUGUI>(true).Where(t => t.name == "Text").First();` initially because SimpleButton prefab doesn't have a TextMeshProUGUI by default under "Content" GameObject so I couldn't query for it as a component.
**Note 2**: The reason why SimpleButton's Content doesn't have children is Mesh's MasterButtonRoot (the source of inspiration for this SimpleButton) doesn't have children under its Content GameObject either.

ms-RistoRK:
> Can you please share some screen shots of the new scenes?

Sure!  Here is one taken from CanvasExampleSimpleButton scene in Editor:
image

org.mixedrealitytoolkit.uxcomponents/Editor/CreateElementMenus.cs
```diff
@@ -210,6 +214,39 @@ private static void CreateActionButton(MenuCommand menuCommand)
             CreateElementFromPath(ActionButtonPath, menuCommand);
         }
 
+        [MenuItem("GameObject/UI/MRTK/Simple Button")]
```
AMollis, line 217: please add an experimental string here

"Simple Button (Experimental)"

org.mixedrealitytoolkit.uxcomponents/Button/Prefabs/SimpleButton.prefab
```diff
@@ -0,0 +1,1172 @@
+%YAML 1.1
```
AMollis, line 1: Please put all new files (*.prefab, *.mat,  *.scene) under an Experimental folder

For example "Button\Prefabs\Experimental\SimpleButton.prefab"

AMollis (changes requested):
I think you may need to update the patch versions for 'standardassets' and 'uxcomponents' packages.  And also update 'uxcomponents's' dependency to use the new 'standardassets' patch.  You update these things in the `package.json` files.

Please refer to Scott's latest release, to determine if package numbers need updating.

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/releases/tag/untagged-e10d3eec6261927bca43

org.mixedrealitytoolkit.uxcomponents/Editor/CreateElementMenus.cs
ms-RistoRK: Done

org.mixedrealitytoolkit.uxcomponents/Button/Prefabs/SimpleButton.prefab
ms-RistoRK: done

ms-RistoRK:
> 
> Also the fonts in the screenshot looks off. Can you make sure it's using the same font and size as the "Action Button"

Re: Fonts size, actually, that is a feature I was trying to add.  The feature is that the Font Size would auto-adjust based on the containers height and always be 25% of the container height so that the initial label ("Simple\nButton") would always have 25% of the container's height as padding when a SimpleButton is created via Editor floating menu, as in:
```
        [MenuItem("GameObject/UI/MRTK/Simple Button (Experimental)")]
        private static void CreateSimpleButton(MenuCommand menuCommand)
        {
            // ...
            textMeshProUGUI.fontSize = simpleButtonContentGameObject.GetComponent<RectTransform>().rect.height / 4;
            // ...
        }
```
... although I notice no other UIElement does font size auto-adjustment so I removed it in the last commit and made all SimpleButtons' Font size the same as ActionButtons'.
The scene now looks like this:
image
I'll work in updating the package references next.

ms-RistoRK:
> 
> It looks like 'standardassets' version needs to be updated from 3.0.2 to 3.1.0; edit 'standassets' package.json and CHANGELOG.md.
> 
Done in latest commit.

org.mixedrealitytoolkit.standardassets/CHANGELOG.md
```diff
@@ -2,8 +2,8 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## [3.0.2-development] - 2024-02-04
+## [3.1.0-development] - 2024-02-09
 
 ### Fixed
 
-* Fixed support for UPM package publishing in the Unity Asset Store.
```
AMollis, line 9: Please keep this line

`* Fixed support for UPM package publishing in the Unity Asset Store.`

Never remove stuff from CHANGELOG.MD, only add.

I just added CHANGELOG's for the Unity Asset Store support, that is why it is pretty bare

AMollis (changes requested):

AMollis:
> > Also the fonts in the screenshot looks off. Can you make sure it's using the same font and size as the "Action Button"
> 
> Re: Fonts size, actually, that is a feature I was trying to add. The feature is that the Font Size would auto-adjust based on the containers height and always be 25% of the container height so that the initial label ("Simple\nButton") would always have 25% of the container's height as padding when a SimpleButton is created via Editor floating menu, as in:
> 
> ```
>         [MenuItem("GameObject/UI/MRTK/Simple Button (Experimental)")]
>         private static void CreateSimpleButton(MenuCommand menuCommand)
>         {
>             // ...
>             textMeshProUGUI.fontSize = simpleButtonContentGameObject.GetComponent<RectTransform>().rect.height / 4;
>             // ...
>         }
> ```
> 
> ... although I notice no other UIElement does font size auto-adjustment so I removed it in the last commit and made all SimpleButtons' Font size the same as ActionButtons'. The scene now looks like this: image I'll work in updating the package references next.

Yikes that font size does not look great.  I suggest we change the default label to "Label", and use the same font settings has the Action Button...does this help?

image

org.mixedrealitytoolkit.standardassets/CHANGELOG.md
ms-RistoRK: Done

ms-RistoRK:
The updated demo scene (CanvasExampleSimpleActionButton) now looks like this:
image

AMollis (approved):
Please add an "MRTK Issue" to add "StateVisualizer" to the new "Simplified Button"

PR #636, opened 2024-02-08
Fix the fallback controllers being backwards

keveleigh:
#559 introduced a regression where the controller models' rotational offset wasn't properly committed to the PR. This PR fixes that.


AMollis (approved):

AMollis:
@shaynie  I think this mean you need to branch and cherry pick this change.  Or we restart the release

PR #638, opened 2024-02-08
Updating Documentation Binaries for Correct Minor Versions

AMollis:
# Pull Request Description: Updating Documentation Binaries for Correct Minor Versions

## Overview 

The current issue lies in the incorrect minor versions of our documentation binaries. Specifically, the `AssemblyInfo.cs` files are not being updated with the latest major/minor version during the documentation build pipeline. To address this, we propose applying the same versioning logic used for package creation to the documentation generation process.

Here are the key changes:

1. **Consolidating Versioning Logic:** We’ve consolidated the versioning logic from both `pack-upm.ps1` and `repackage-for-release.ps1` into a single PowerShell script named `update-versions.ps1`. This script can handle versioning for both prerelease and release packages, following the current version patterns (semantic versioning with labels and appended revision numbers as needed). The goal here is to not change the versioning logic. In the future we may want to revisit this. For example, in a future pull request, we might want to remove the usage of the `pre.x` tags on experimental packages, where `x` is an additional experimentation revision number shared across all package. This experimental tag adds additional complexity to the build scripts, and might be avoided if we just used a `-prerelease` label with the already utilized semantic versioning.
3. **Handling Prerelease to Release Transition:** The new script is also capable of re-versioning a prerelease package to a release package seamlessly.
4. **Integration with `doc.yaml`**: We’ve integrated the new script into `doc.yaml`. Now, when generating documentation binaries, the script ensures that they appear as release packages. This adjustment allows for correct moniker (i.e., directory) names to be created. That is, when a moniker is created for a package, the external doc generation script queries the major and minor versions of the doc libraries (DLL).
5. **Improved Version Consistency:** Previously, the doc generation script relied on hard-coded major and minor versions, which didn’t always match the versions specified in the packages’ `package.json` files. With this change, we maintain a single source of versioning truth—the `package.json`—and expect the build pipeline to propagate the version to other relevant areas, including the AssemblyInfo.cs files.

By implementing these modifications, we aim to enhance the accuracy and consistency of our documentation binaries. 

## Fixes
- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/637

PR #639, opened 2024-02-09
Manually update versions in  AssemblyInfo files for the release branch

shaynie:
Manually update versions in  AssemblyInfo files for the release branch.  This is a manual fix for generating the docs.

### DIFF
--- a/org.mixedrealitytoolkit.accessibility/AssemblyInfo.cs
+++ b/org.mixedrealitytoolkit.accessibility/AssemblyInfo.cs
@@ -12,4 +12,4 @@ using System.Runtime.CompilerServices;
 // The AssemblyVersion attribute is checked-in and is recommended not to be changed often.
 // https://docs.microsoft.com/troubleshoot/visualstudio/general/assembly-version-assembly-file-version
 // AssemblyFileVersion and AssemblyInformationalVersion are added by pack-upm.ps1 to match the current MRTK build version.
-[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyVersion("1.0.1.0")]
--- a/org.mixedrealitytoolkit.audio/AssemblyInfo.cs
+++ b/org.mixedrealitytoolkit.audio/AssemblyInfo.cs
@@ -9,4 +9,4 @@ using System.Reflection;
 // The AssemblyVersion attribute is checked-in and is recommended not to be changed often.
 // https://docs.microsoft.com/troubleshoot/visualstudio/general/assembly-version-assembly-file-version
 // AssemblyFileVersion and AssemblyInformationalVersion are added by pack-upm.ps1 to match the current MRTK build version.
-[assembly: AssemblyVersion("3.0.0.0")]
+[assembly: AssemblyVersion("3.0.1.0")]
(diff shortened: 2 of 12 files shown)


org.mixedrealitytoolkit.accessibility/AssemblyInfo.cs
```diff
@@ -12,4 +12,4 @@
 // The AssemblyVersion attribute is checked-in and is recommended not to be changed often.
 // https://docs.microsoft.com/troubleshoot/visualstudio/general/assembly-version-assembly-file-version
 // AssemblyFileVersion and AssemblyInformationalVersion are added by pack-upm.ps1 to match the current MRTK build version.
-[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyVersion("1.0.1.0")]
```
keveleigh, line 15: Oh hmm, these aren't the attributes that are changed when pack-upm.ps1 is run (or at least, they didn't used to be!). The comment above this line explains why, but I'm happy to discuss.

keveleigh: Oh no, it looks like this changed after I wrote the original packing script 😭 it was intentional that this entry was checked-in and not changed often

https://github.com/microsoft/MixedRealityToolkit-Unity/commit/9c4a63cbf175c37634ab5bc5101381a8c57f987e#diff-88f30405f832e564adece2412b83cf181d8867568037debe7702f2b8dabbd821R122

image

keveleigh: It probably doesn't matter as much, since we aren't shipping precompiled binaries ¯\_(ツ)_/¯ idk, it's probably fine haha

org.mixedrealitytoolkit.audio/AssemblyInfo.cs
```diff
@@ -9,4 +9,4 @@
 // The AssemblyVersion attribute is checked-in and is recommended not to be changed often.
 // https://docs.microsoft.com/troubleshoot/visualstudio/general/assembly-version-assembly-file-version
 // AssemblyFileVersion and AssemblyInformationalVersion are added by pack-upm.ps1 to match the current MRTK build version.
-[assembly: AssemblyVersion("3.0.0.0")]
+[assembly: AssemblyVersion("3.0.1.0")]
```
keveleigh, line 12: > This is a manual fix for generating the docs.

Is this a temporary manual fix? Or something we're going to need to go going forward?

AMollis: > This is a temporary fix only needed for this release, so docs binaries have the correct versions.   We have fix this correct for main in this pr 
> 
> https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/638

org.mixedrealitytoolkit.accessibility/AssemblyInfo.cs
AMollis: This is a temporary fix only needed for this release, so docs binaries have the correct versions.   We have fix this correct for main in this pr 

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/638

AMollis (approved):

org.mixedrealitytoolkit.audio/AssemblyInfo.cs
keveleigh: Excellent! Thanks for clarifying

AMollis:
@keveleigh , I'm pushing this through so I can generate docs.  Hopefully you're ok with this for ONLY this release.

This is a temporary fix only needed for this release, so the doc binaries have the correct versions.   We have this fixed correctly in the main branch.  

- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/638

The doc pipeline does not update the assembly versions. We need correct versions, so the monikers (i.e. directory names) in the build output are correct.

PR #648, opened 2024-02-16
Fixes buttons of secondary keyboard layouts not making any sound

LocalJoost:
The NonNativeKeyboard consists out of several panes. the keyboard_Alpha and keyboard_Space_Alpha are active. However, if you press the &123 button, the keyboard_Symbols gets active. The buttons on this pane do not make sound, because it uses `GetComponentsInChildren<Button>()` to find the button, which only finds active game objects, while it should use `GetComponentsInChildren<Button>(true)`, which finds both active _and_ inactive game objects


marlenaklein-msft (approved):

marlenaklein-msft:
thanks for the fix!

LocalJoost:
You are welcome!

PR #649, opened 2024-02-16
Easier way to utilize more performant button (with dynamic front plate)

ms-RistoRK:
* Added Empty, Action, Action with Checkbox, and ToggleSwitch Experimental buttons prefabs
* Added corresponding MenuItems for use in Editor Tools
* Added CanvasUITearsheetDynamicFrontplate scene


org.mixedrealitytoolkit.uxcore/Button/Experimental/ExperimentalPressableButtonWithDynamicFrontPlate.cs
```diff
@@ -0,0 +1,79 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace MixedReality.Toolkit.UX
+{
+    /// <summary>
+    /// This mono behaviour is used to enable or disable the front plate of a button when the button enters or exists proximity-hovering.
+    /// </summary>
+    [RequireComponent(typeof(PressableButton))]
+    public class ExperimentalPressableButtonWithDynamicFrontPlate : MonoBehaviour
```
AMollis, line 13: Remove the name "Experimental"  and add the following remarks XML.  

```cs
    /// <remarks>
    /// This is an experimental feature. This class is early in the cycle, it has 
    /// been labeled as experimental to indicate that it is still evolving, and 
    /// subject to change over time. Parts of the MRTK, such as this class, appear 
    /// to have a lot of value even if the details haven't fully been fleshed out. 
    /// For these types of features, we want the community to see them and get 
    /// value out of them early enough so to provide feedback. 
    /// </remarks>
```

org.mixedrealitytoolkit.uxcore/Button/Experimental/ExperimentalPressableButtonWithDynamicFrontPlate.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+            return null;
+        }
+
+        #endregion Private Members
+
+        /// <summary>
+        /// A Unity event function that is called on the frame when a script is enabled just before any of the update methods are called the first time.
+        /// </summary> 
+        private void Start()
```
AMollis, line 52: As a best practice we always set this "Unity Event Functions" to "protected"  this way the user will get a compiler error if they derive and try to make their own Start() method.


If we left as private when something derives from this and creates its Start() method, the derived class's Start() method would hide the base one.  SO making this `projected` would prevent this error.

org.mixedrealitytoolkit.uxcore/Button/Experimental/ExperimentalPressableButtonWithDynamicFrontPlate.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        #region Private Members
+
+        /// <summary>
+        /// Gets the front plate's raw image component if this is an EmptyButton (Experimental), ActionButton (Experimental), or CanvasButtonToggleSwitch (Experimental).  Null otherwise.
+        /// </summary>
+        /// <returns>Reference to this button FrontPlate's RawImage Component.  Null if it doesn't exist or if this is not an EmptyButton (Experimental), ActionButton (Experimental), or CanvasButtonToggleSwitch (Experimental)</returns>
+        internal RawImage GetFrontPlateRawImage()
+        {
+            if (gameObject.tag.Equals("ExperimentalDynamicFrontplate")) //This is a temporary conditional for the experimental dynamic frontplate feature, it will be removed if the community accepts the experimental feature to be included in official release and be integrated as part of PressableButton script
```
AMollis, line 34: I see your comment...however even in experimental components we should avoid poor practices.

Please add a serializable field to this class.

org.mixedrealitytoolkit.uxcore/Button/Experimental/ExperimentalPressableButtonWithDynamicFrontPlate.cs
AMollis, line 13: I don't understand the need for this class.   You've created a base "Empty Button" template.  You 
can hook up then event in the "Empty Button (Experimental)"'s inspector.  No need for code.

AMollis (changes requested):

org.mixedrealitytoolkit.uxcore/Button/Experimental/ExperimentalPressableButtonWithDynamicFrontPlate.cs
ms-RistoRK: Makes sense, thank you for the suggestion, implemented in latest commit.

org.mixedrealitytoolkit.uxcore/Button/Experimental/ExperimentalPressableButtonWithDynamicFrontPlate.cs
ms-RistoRK: I was trying to keep all the Experimental code as separated as possible from the release code so that it would be easy to identify and update as needed.  Also, I added the class so that existing buttons wouldn't have their behavior modified by this experimental code.

org.mixedrealitytoolkit.uxcore/Button/Experimental/ExperimentalPressableButtonWithDynamicFrontPlate.cs
ms-RistoRK: Mmmmmmmmhhhh poor practice? ... I disagree to call my approach a poor-practice because it empowers users to immediately benefit from this change and removes human-error from the equation.  On the other hand, I believe the approach of using a SerializedField to manually set the Frontplate's RawImage is human-error prone and a waste of time for the user of Buttons with Dynamic Frontplates because he/she has to do something that could be done automatically as in my approach.  For example:  To update CanvasUITearsheedDynamicFrontplate I (or any other user) will have to go through all 90-something buttons in the scene and manually drag-and-drop each button's Frontplate's RawImage to make it work.  IMHO that's a waste of time and human-error prone.
Lastly, since I agreed to follow the framework's philosophy then I'll make the change but I think we should keep an open-mind to different approaches that are more useful to our customers in the terms of development speed and reducing human-errors.

ms-RistoRK: Ok, so, I stand corrected.  I talked to Adam offline and I now have a better understanding of this feedback.  His suggestion is to modify the Component property value (in this case the RawImage's Enabled property) directly in the Inspector's event call.  I, wrongfully, was focused on having to do it via a script but now that I better understand the feedback, I'll do it that way.

ms-RistoRK: So ... in last commit I have removed the ExperimentalPressableButtonWithDynamicFrontPlate script and also made the Frontplate's RawImage to be {enabled|disabled} on ProximityHovered{Entered|Exited} events respectively.  This change make the FrontPlateRawImage serializable field no longer necessary.  Resolving this.

org.mixedrealitytoolkit.uxcore/Button/Experimental/ExperimentalPressableButtonWithDynamicFrontPlate.cs
ms-RistoRK: Implemented with event hooking directly in Inspector in latest commit.  Resolving this.

org.mixedrealitytoolkit.uxcomponents/Editor/CreateElementMenus.cs
```diff
@@ -235,10 +251,74 @@ private static void CreateSimpleActionButton(MenuCommand menuCommand)
             Undo.SetCurrentGroupName("Create SimpleActionButton");
 
             GameObject simpleActionButton = CreateElementFromPath(SimpleActionButtonPath, menuCommand);
-
             Undo.RecordObject(simpleActionButton, "Added SimpleActionButton instance.");
         }
 
+        #region Experimental Buttons with dynamic Frontplate
+        [MenuItem("GameObject/UI/MRTK/Dynamic Frontplate (Experimental)/Empty Button w dynamic Frontplate (Experimental)")]
```
AMollis, line 258: To not pollute the menus with experimental stuff at the top, let's move all the experimental stuff under a new submenu "Experimental"

For example, 

"GameObject/UI/MRTK/Experimental/Dynamic Frontplate/Action Button (Wide)"

org.mixedrealitytoolkit.uxcomponents/Editor/CreateElementMenus.cs
```diff
@@ -235,10 +251,74 @@ private static void CreateSimpleActionButton(MenuCommand menuCommand)
...
+        private static void CreateEmptyButtonDynamicFrontplate(MenuCommand menuCommand)
```
AMollis, line 259: Let's not add a menu for the empty button....I don't think that's very useful given there is no visual component.

org.mixedrealitytoolkit.uxcomponents/Editor/CreateElementMenus.cs
AMollis: move the "simple buttons" under the "Experimental" menu as well.

Also, remove the "(Experimental)" labels in the names.  Since the items would be under "Experimental" menu, having "(Experimental)" seem redundant.

AMollis (changes requested):

org.mixedrealitytoolkit.uxcomponents/Editor/CreateElementMenus.cs
ms-RistoRK: Removed in latest commit.  Resolving.

AMollis (dismissed):

AMollis (approved):

PR #655, opened 2024-02-23
Makes non native keyboard touch enabled

LocalJoost:
Makes non native keyboard touch enabled, as described in https://localjoost.github.io/Using-the-non-native-keyboard-in-touch-scenarios-in-MRTK3/

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/CHANGELOG.md
+++ b/org.mixedrealitytoolkit.spatialmanipulation/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Added
 
 * Added different types of maintaining scale for bounds handles. [PR #976](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/976)
-* Added RadialViewDisplayInitializeHelper to position a game object with a RadialView component at the center of the view. [PR #655](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/655)
+* Added InitializeToTargetForward field/property to RadialView to position a GameObject in front of its target (like at the center of the view) when it becomes enabled. [PR #655](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/655)
 
 ### Fixed
 


org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
+﻿// Copyright (c) Mixed Reality Toolkit Contributors
+// Licensed under the BSD 3-Clause
+using System.Collections;
```
AMollis, line 3: Nit.  Add new line

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+        private Graphic image;
+        private float lastClickTime;
+        private bool isInitialized;
+        private Vector3 defaultPosition;
+        private Vector3 animatedPosition;
+        private BoxCollider buttonCollider;
+        private Vector3 buttonColliderDefaultCenter;
+        private Color defaultImageColor;
+
+        private void Awake()
```
AMollis, line 31: As a best practive we are requiring that MonoBehaviour event functions be at least "protected", if the class is not sealed.

This requirement is meant to prevent subclasses from accidentally hiding the base's implementation of event functions (Awake, OnEnable, ect.).  That is, if the base event functions are private and the subclass implemented its own version, the compiler will happily allow this (no warning, no error).   This would then hide the base implementation. 

If the base event functions are protected, the compiler will at least provide a warning message.

So please make `Awake` , `OnEnable`, and `Update` projected.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+using System.Collections;
+using MixedReality.Toolkit;
+using MixedReality.Toolkit.Input;
+using UnityEngine;
+using UnityEngine.UI;
+using UnityEngine.XR.Interaction.Toolkit;
+
+namespace MixedReality.Toolkit.UX.Experimental
+{
+    public class NonNativeKeyTouchAdapter : MonoBehaviour
```
AMollis, line 12: All public, protected types and members need to be documented with a `<summary>` tag.   Please add a `<summary>` tag,

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        private void Initialize()
+        {
+            if (isInitialized)
+            {
+                return;
+            }
+            isInitialized = true;
+
+            var rectTransform = GetComponent<RectTransform>();
```
AMollis, line 57: Add "Require Component" attribute to this class.

`[RequireComponent(typeof(RectTransform))]`

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+            buttonCollider = gameObject.EnsureComponent<BoxCollider>();
```
AMollis, line 58: I don't think we should be automatically creating a box collider here.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
AMollis, line 57: I prefer having the caller explicitly set the component values via serialized fields, instead of using `GetComponent`.    Doing so will allow this component to be utilized by various other hierarchies.  This would also remove the need for "required components" on this game object.

Curious what other reviewers think  (@keveleigh ?)

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+            buttonCollider.size = size;
+            buttonColliderDefaultCenter = new Vector3((size.x + ColliderMargin) / 2.0f,
+                (-size.y - ColliderMargin) / 2.0f, ColliderZDelta);
+            buttonCollider.center = buttonColliderDefaultCenter;
+            
+            image = GetComponent<Graphic>();
+            defaultImageColor = image.color;
+            var button = GetComponent<Button>();
+            
+            interactable = gameObject.EnsureComponent<StatefulInteractable>();
```
AMollis, line 72: Please remover "EnsureComponent"   Require stateful Interactable instead.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+            var size = new Vector3(
+                rectTransform.rect.size.x - ColliderMargin,
+                rectTransform.rect.size.y - ColliderMargin, 
+                ColliderThickness);
+            buttonCollider.size = size;
+            buttonColliderDefaultCenter = new Vector3((size.x + ColliderMargin) / 2.0f,
+                (-size.y - ColliderMargin) / 2.0f, ColliderZDelta);
+            buttonCollider.center = buttonColliderDefaultCenter;
+            
+            image = GetComponent<Graphic>();
```
AMollis, line 68: Require graphic

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+            defaultImageColor = image.color;
+            var button = GetComponent<Button>();
```
AMollis, line 70: Require button

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+            if (interaction is PokeInteractor)
+            {
+                image.color = color;
+            }
+        }
+        
+        private void Update()
+        {
+            // fallback for if the user very rapidly moves their finger over the keyboard and
+            // interactable.lastHoverExited does not get fired (or too early)
```
AMollis, line 113: Seems like you are working around a bug here.   Is there something else we can fix to prevent this component from "missing" the lastHoverExited event?

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
...
+                i += Time.deltaTime * rate;
+                var newPos = Vector3.Lerp(startPos, endPos, Mathf.SmoothStep(0f, 1f, i));
+                transform.localPosition = newPos; buttonCollider.center = buttonColliderDefaultCenter - (newPos - defaultPosition);
+                yield return null;
+            }
+        }
+
+        private void OnDestroy()
+        {
+            interactable.hoverExited.RemoveAllListeners();
```
AMollis, line 139: Please only remove the explicit handlers you added.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyboardTouchAdapter.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using Microsoft.MixedReality.GraphicsTools;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace MixedReality.Toolkit.UX.Experimental
+{
+    public class NonNativeKeyboardTouchAdapter : MonoBehaviour
+    {
+        private void Awake()
```
AMollis, line 12: As a best practive we are requiring that MonoBehaviour event functions be at least "protected", if the class is not sealed.

This requirement is meant to prevent subclasses from accidentally hiding the base's implementation of event functions (Awake, OnEnable, ect.).  That is, if the base event functions are private and the subclass implemented its own version, the compiler will happily allow this (no warning, no error).   This would then hide the base implementation. 

If the base event functions are protected, the compiler will at least provide a warning message.

So please make `Awake` , `OnEnable`, and `Update` projected.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyboardTouchAdapter.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Mixed Reality Toolkit Contributors
+// Licensed under the BSD 3-Clauseusing System;
+
+using Microsoft.MixedReality.GraphicsTools;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace MixedReality.Toolkit.UX.Experimental
+{
+    public class NonNativeKeyboardTouchAdapter : MonoBehaviour
```
AMollis, line 10: All public, protected types and members need to be documented with a `<summary>` tag.   Please add a `<summary>` tag,

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyboardTouchAdapter.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private void Awake()
+        {
+            var defaultAudioComponent = GetComponent<AudioSource>();
```
AMollis, line 14: Add required component.  

I prefer a "serialized field" however.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyboardTouchAdapter.cs
```diff
@@ -0,0 +1,32 @@
...
+            defaultAudioComponent.spatialize = false;
+        }
+
+        private void Start()
+        {
+            var buttons = GetComponentsInChildren<Button>(true);
+            foreach (var button in buttons)
+            {
+                // The search box has an incorrect collider and should not act as a button anyway
+                if (button.gameObject.name != "search")
```
AMollis, line 25: This seems a bit hacky...querying for "search".

Why can't you explicitly add the "NonNativeKeyTouchAdapter" to the buttons in the keyboard?  That would remove the need for this class.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyboardTouchAdapter.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace MixedReality.Toolkit.UX.Experimental
+{
+    public class NonNativeKeyboardTouchAdapter : MonoBehaviour
+    {
+        private void Awake()
+        {
+            var defaultAudioComponent = GetComponent<AudioSource>();
+            defaultAudioComponent.playOnAwake = false;
+            defaultAudioComponent.spatialize = false;
```
AMollis, line 16: I don't think we should be explicitly disabled spatial audio here.   Why should allow the prefab to define this.

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
```diff
@@ -0,0 +1,27 @@
+using UnityEngine;
```
AMollis, line 1: Missing Copyright  header

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace MixedReality.Toolkit.SpatialManipulation
+{
+    /// <summary>
+    /// Makes a game object appear having a RadialView component in the center of the view,
+    /// at a distance equal to the average of the MinDistance and MaxDistance of the RadialView.
+    /// This prevents the object from 'racing into your view' and/or appear slanted
```
AMollis, line 8: remove single questions around "racing into you view"

AMollis: Try this instead.

```cs
/// <summary>
/// Positions a game object with a RadialView component at the center of the view,
/// maintaining a distance equal to the average of the MinDistance and MaxDistance 
/// properties of the RadialView. This prevents the object from rapidly approaching the 
/// viewer and ensures it appears upright.
/// </summary>
```

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
```diff
@@ -0,0 +1,27 @@
...
+    public class RadialViewDisplayInitializeHelper : MonoBehaviour
+    {
+        private RadialView radialView;
+
+        protected void Awake()
+        {
+            radialView = GetComponent<RadialView>();
+        }
+
+        private void OnEnable()
```
AMollis, line 20: As a best practive we are requiring that MonoBehaviour event functions be at least "protected", if the class is not sealed.

This requirement is meant to prevent subclasses from accidentally hiding the base's implementation of event functions (Awake, OnEnable, ect.).  That is, if the base event functions are private and the subclass implemented its own version, the compiler will happily allow this (no warning, no error).   This would then hide the base implementation. 

If the base event functions are protected, the compiler will at least provide a warning message.

So please make  `OnEnable` projected.

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
AMollis, line 1: Please add change comment to "Spatial Manipulation's" CHANGELOG.md.   Following the [Keep A Changelog](https://keepachangelog.com/en/1.1.0/) pattern.

Also please change that "Spatial Manipulation's version has been properly incremented since last [release](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/releases).

org.mixedrealitytoolkit.uxcomponents/Experimental/NonNativeKeyboard/TouchableNonNativeKeyboard.prefab
```diff
@@ -0,0 +1,174 @@
+%YAML 1.1
```
AMollis, line 1: Please add change comment to "UX Component's" CHANGELOG.md.   Following the [Keep A Changelog](https://keepachangelog.com/en/1.1.0/) pattern.

Also please change that "UX Component's version has been properly incremented since last [release](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/releases).

Also please make sure "UX Component's" dependency versions have been properly incremented. Since you are depending on new feature in "UX Core"

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
```diff
@@ -0,0 +1,145 @@
+﻿// Copyright (c) Mixed Reality Toolkit Contributors
```
AMollis, line 1: Please add change comment to "UX Core's" CHANGELOG.md.   Following the [Keep A Changelog](https://keepachangelog.com/en/1.1.0/) pattern.

Also please change that "UX Core's" version has been properly incremented since last [release](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/releases).

AMollis (changes requested):
Thank for you changes.  I'm made some change requests.

@keveleigh  and @marlenaklein-msft ... can you please review as well.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: Uhm... I think we do. Otherwise, how will it be touchable otherwise?

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyboardTouchAdapter.cs
LocalJoost: The buttons are not a prefab. In theory I an do what you want. Manually add them, one by one, for every key. I thought the code solution was pretty neat

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: Done

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: Done

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: Done

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: Done

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: Done

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyboardTouchAdapter.cs
LocalJoost: Done

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyboardTouchAdapter.cs
LocalJoost: Done

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
LocalJoost: Done

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
LocalJoost: Done. Thanks for this ready-to-copy header :)

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: The whole idea is to create an add-on for the existing keyboard. This behaviour is added in code (by NonNativeKeyboardTouchAdapter) and isn't actually indented to be used on it's own

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyboardTouchAdapter.cs
LocalJoost: Done

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: Refactored anonymous lambda's into separate methods that can be individually removed. However - this _is_ the OnDestroy method, right? Nothing much happens after this ;)

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: I think I am working around a bug indeed, but I fear that might be in StatefulInteractable. This typically happens when you bring your finger at 'hover distance' from a button, let's say Q, and then rapidly move your finger over the keyboard to the right. It looks like events are then fired in the wrong order. I changed my code to not use the lastHoverExited at all.

LocalJoost:
I don't understand "please link to an issue". There is no issue, this is more or less new functionality. Or actually, functionality that got lost when the NonNativeKeyboard got ported from MRTK2. As to Unit tests... that is going to be tricky. The keyboard as such still runs as before. I will have a look

org.mixedrealitytoolkit.uxcomponents/Experimental/NonNativeKeyboard/TouchableNonNativeKeyboard.prefab
LocalJoost: I have updated the ChangeLog but I have no idea what you mean by "Also please change ... " What am I supposed to do? Edit the asmdef?

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var distance = (radialView.MinDistance + radialView.MaxDistance) / 2.0f;
```
marlenaklein-msft, line 26: I wonder if it would make more sense to include this as a configurable behavior on RadialView instead of adding an additional script. I.E. adding a serialized field CenterInitialPosition, and if that is true then making this adjustment.

That would also allow SolverHandler.TransformTarget to be used in place of Camera.main.transform

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
LocalJoost: I have updated the ChangeLog but I have no idea what you mean by ""Spatial Manipulation's version has been properly incremented since last release" What am I supposed to do? Edit the asmdef?

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
LocalJoost: That is a possibility. I was trying to mess as little as possible with existing code. :)

marlenaklein-msft: @AMollis do you have a preference here?

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
marlenaklein-msft: I think this would be updating the package.json to "version": "3.2.0-development"

Since this adds a new public class it will require a minor version bump, and the most recent release was 3.1.0. Nobody has bumped the minor version since then so this change would need to

org.mixedrealitytoolkit.uxcomponents/Experimental/NonNativeKeyboard/TouchableNonNativeKeyboard.prefab
marlenaklein-msft: I believe this would be updating the package.json to "version": "3.2.0-development". Last version of UX Core was 3.1.1

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
marlenaklein-msft: This should also be "version": "3.2.0-development". Thank you!

org.mixedrealitytoolkit.uxcomponents/Experimental/NonNativeKeyboard/TouchableNonNativeKeyboard.prefab
marlenaklein-msft: The dependencies in the package.json file will also need to be updated, since this depends on the spatial manipulation and uxcore changes you are making

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: This is already 3.2.0-development!

org.mixedrealitytoolkit.uxcomponents/Experimental/NonNativeKeyboard/TouchableNonNativeKeyboard.prefab
LocalJoost: I think I fixed what you asked. I hope I did it correctly

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
LocalJoost: I think I fixed what you ask, I hope I did it correctly

LocalJoost:
@AMollis @marlenaklein-msft so I addressed all the points I think should be addressed, and added two unit test, one actually testing a hand pressing the "j" actually resulting in a "j" appearing in the keyboard preview. I hope this satisfies your needs.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
AMollis: let's just call "get component"  I would like the prefab to explicitly define their own collider.   Let's follow the pattern used by XRI.   Perfab can specify colliders via serialized field, if not specific call get components in children

Example from XRBaseInteractable
```cs

        /// <summary>
        /// See <see cref="MonoBehaviour"/>.
        /// </summary>
        protected virtual void Awake()
        {
            // If no colliders were set, populate with children colliders
            if (m_Colliders.Count == 0)
            {
                GetComponentsInChildren(m_Colliders);
                // Skip any that are trigger colliders since these are usually associated with snap volumes.
                // If a user wants to use a trigger collider, they must serialize the reference manually.
                m_Colliders.RemoveAll(col => col.isTrigger);
            }
```

AMollis (changes requested):
I thought about this some more.  Is building a "touch adapter" the right thing here?

Back when we added this component, we decided to first release a "flat canvas keyboard" similar to the old MRTK2 one, so we had a starting point.  We would then latter update the keyboard with "pressable buttons".

That is, instead of added a touch adapter, can we just use XRIteractables instead.  Like PressableButton?

LocalJoost:
> I thought about this some more. Is building a "touch adapter" the right thing here?
> 
> Back when we added this component, we decided to first release a "flat canvas keyboard" similar to the old MRTK2 one, so we had a starting point. We would then later update the keyboard with "pressable buttons".
> 
> That is, instead of creating and adding a touch adapter, can we just use XRInteractables instead. Like PressableButton?
> 
> Sidenote, the XRI's [poke interactor](https://docs.unity3d.com/Packages/com.unity.xr.interaction.toolkit@2.3/manual/xr-poke-interactor.html) works with UGUI canvas elements. That is, XRI's poke interactor can "touch" UGUI buttons. Ideally MRTK3's poke interactor would also work with UGUI buttons, however doing so requires access to an internal XRI type `UIInteractionType`.
> 
> ```cs
>     /// <summary>
>     /// Defines the type of interaction expected from the interactor when interacting with UI.
>     /// </summary>
>     /// <seealso cref="TrackedDeviceModel.interactionType"/>
>     internal enum UIInteractionType
>     {
>         /// <summary>
>         /// The UI interaction is a ray interaction.
>         /// </summary>
>         Ray,
> 
>         /// <summary>
>         /// The UI interaction is a poke interaction.
>         /// </summary>
>         Poke,
>     }
> ```
> 
> In the next major version, I'd like to get rid of MRTK3's Poke Pointer, and use XRI's. This would unlock UGUI touch interactions

This is a bit confusing to me. It was my intention to create a simple, small add-on to make the existing non-native keyboard work with HoloLens 2 as it is now not very usable, or at least clashing quite heavily with the usage pattern. And I needed it for my own app, so I made it to work and thought it nice to add it so the MRTK3 itself. It seems now that you want to go in a complete different direction, or maybe even two ;) Do you rather want a redesign of the keyboard? Am I correct in understanding that you don't want to merge this at all? I can release it as an add-on package and keep it outside of the MRTK3, if you think that is a better idea.

AMollis:
> of the keyboard? Am I correct in understanding that you don't want to merge this at all? I can release it as an add-on package and keep i

Two keyboard options.

1. Use UGUI Buttons
2. Use MRTK "Pressable" Buttons

It sounds like you prefer #1.   While I prefer #2.  If we wanted to continue with #1, I would much prefer getting the MRTK Poke Interactor to work with all UGUI Buttons, without the need of colliders, similar to how the XRI Poke Interactor.  

I've pinged the XRI folks to see how we can make MRTK's poke interactor work with UGUI Buttons.   I'll follow-up here....

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: I will repeat: I can't change anything to the _prefab of a key_ because _a key is not a prefab_. All keys are just gameobjects in the keyboard. See? No blue cubes
image

So I have to add every piece of infrastructure myself. And if I, per your suggestion, do not create the collider, the result it the touch no longer works. The unit test fails. 

image


Please try it yourself. I have uploaded, for your convenience, a version of NonNativeKeyTouchAdapter.cs that has all the collider stuff deleted. 
[NonNativeKeyTouchAdapter.zip](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/files/14513655/NonNativeKeyTouchAdapter.zip)

LocalJoost:
> > of the keyboard? Am I correct in understanding that you don't want to merge this at all? I can release it as an add-on package and keep i
> 
> Two keyboard options.
> 
> 1. Use UGUI Buttons
> 2. Use MRTK "Pressable" Buttons
> 
> It sounds like you prefer 1. While I prefer 2. If we wanted to continue with 1, I would much prefer getting the MRTK Poke Interactor to work with all UGUI Buttons, without the need of colliders, similar to how the XRI Poke Interactor works.
> 
> I've pinged the XRI folks to see how we can make MRTK's poke interactor work with UGUI Buttons. I'll follow-up here....

I don't prefer so much one or the other, I prefer something that works, without much work, and that works _now._ Yes, I can rebuild the keyboard using actual MRTK3 PressableButtons (which are, incidentally child classes of the StatefulInteractable that I use). But then I would have to add all the graphics part as well - so, in fact, replace all the entries in the keyboard with Action Button prefabs. Manually as, the keys are now no prefabs, I would basically have to build the whole keyboard from the ground up. Key by key. Definitely doable, but a lot of work for something that I don't think a lot of people use. So I took a piece of code to do the trick. WAY less work and leaves current functionality intact. 

And the Action Button, incidentally, uses a BoxCollider as well. Where do you think I got the idea in the first place? Great artists steal ;) Also, apparently the alternative way you suggest requires a change to an external (Unity) package while mine just uses what is readily available. Today.

org.mixedrealitytoolkit.uxcore/Experimental/NonNativeKeyboard/NonNativeKeyTouchAdapter.cs
LocalJoost: The only way to get this done is to add, this behaviour, manually, one by one to _every_ key in the keyboard - as the keys are no prefabs. I feel like a broken record repeating this ;)

LocalJoost:
So is there already a version of MRTK released that makes the native keyboard touch enabled? 🙄

org.mixedrealitytoolkit.spatialmanipulation/Solvers/RadialViewDisplayInitializeHelper.cs
keveleigh: I went ahead and migrated this behavior into RadialView, blocked behind a new `InitializeToTargetForward` flag (which is enabled on the touchable keyboard prefab).

keveleigh (approved):

LocalJoost:
💪💪

PR #611, opened 2024-01-10
Adding IsProximityHovered property of type TimedFlag to detect when a button starts being hovered or on interactor proximity and when it stops being hovered or on proximity of any interactor.

ms-RistoRK:
This PR is to adds IsProximityHovered property of type TimedFlag to detect when a button starts being hovered or on interactor proximity and when it stops being hovered or on proximity of any interactor.
Such property will be used in the near future to make the FrontPlate's RawImage to be dynamically enabled/disabled based on Proximity + Hovering Enter/Exit events.

* The motivation of such new behavior is the research done to find bottlenecks on MRTK3 buttons and increase the amount of concurrent buttons in a scene before a significant framerate drop (60 FPS).
* The impact of this change is that MRTK3 scenes will now be able to support ~84% of buttons as in MRTK2 scenes (current is ~46%), effectively doubling the amount of buttons before a significant framerate drop (also documented).
* The philosophy behind this code change is:
  * We retain the investment previously done on the improved visual appeal of MRTK3 buttons, nothing is erased.
* New UnityTests for the above ^ are also included in this PR.

PR #659, opened 2024-02-26
Additional feedback for pr611

ms-RistoRK:
This PR addresses the additional feedback kindly provided in [PR 611](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/611) after it was merged.
@keveleigh  Could you please take a look at this PR?


org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -25,49 +28,68 @@ public class NearInteractionModeDetector : ProximityDetector
...
+            }
+        }
+
+        /// <summary>
+        /// Calll OnProximityEntered for all colliders that are currently detected but were not detected previously.  This
+        /// effectively triggers OnProximityEntered for all colliders that are newly detected.
+        /// </summary>
+        private void UpdateProximityEntered()
+        {
+
```
AMollis, line 73: nit.  remove extra new line

AMollis (approved):
Looks good.   Please obtain sign-off from @keveleigh before merging

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Whoops, my bad.  Removed in latest commit, resolving.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -25,49 +28,67 @@ public class NearInteractionModeDetector : ProximityDetector
...
+                        xrProximityInteractable.OnProximityExited(new ProximityExitedEventArgs(previouslyDetectedCollider, xrBaseInteractor));
                     }
+                    previouslyDetectedColliders.Remove(previouslyDetectedCollider);
                 }
-                foreach (Collider collider in DetectedColliders)
+            }
+        }
+
+        /// <summary>
+        /// Calll OnProximityEntered for all colliders that are currently detected but were not detected previously.  This
```
keveleigh, line 68:
```suggestion
        /// Call OnProximityEntered for all colliders that are currently detected but were not detected previously.  This
```

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -25,49 +28,67 @@ public class NearInteractionModeDetector : ProximityDetector
...
-                        {
-                            foreach (XRBaseInteractor xrBaseInteractor in nearInteractors)
-                            {
-                                previouslyDetectedCollider.GetComponent<IXRProximityInteractable>().OnProximityExited(new ProximityExitedEventArgs(this, previouslyDetectedCollider, xrBaseInteractor));
-                            }
-                        }
-                        previouslyDetectedColliders.Remove(previouslyDetectedCollider);
+                        xrProximityInteractable.OnProximityExited(new ProximityExitedEventArgs(previouslyDetectedCollider, xrBaseInteractor));
                     }
+                    previouslyDetectedColliders.Remove(previouslyDetectedCollider);
```
keveleigh, line 62: I still really, really don't think we should be removing from `previouslyDetectedColliders` while we're actively iterating through it. I don't see https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/611#discussion_r1489859824 addressed here, but maybe I'm misreading the change. (I think my previous comment may have been posted on the wrong loop)

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -25,49 +28,67 @@ public class NearInteractionModeDetector : ProximityDetector
...
+        private void UpdateProximityEntered()
+        {
+            foreach (Collider collider in DetectedColliders)
+            {
+                if (!previouslyDetectedColliders.Contains(collider) &&
+                    InteractionManager.TryGetInteractableForCollider(collider, out IXRInteractable xrInteractable) &&
+                    xrInteractable is IXRProximityInteractable xrProximityInteractable)
                 {
-                    if (!previouslyDetectedColliders.Contains(collider))
+                    foreach (XRBaseInteractor xrBaseInteractor in nearInteractors)
```
keveleigh, line 79: I'm also still curious about your thoughts on https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/611#discussion_r1489878783

> Instead of sending multiple events to account for all `nearInteractors`, should we just send _one_ relaying info about either:
> 
> 1. The `ProximityDetector` itself. We could potentially expose the list of `nearInteractors` as an `IReadOnlyList` or something if we think the interactables need that info
> 2. Specifically the `nearInteractor`(s) doing the selecting. They likely aren't all selecting at once, but one of the cases for this logic is that _any_ `nearInteractor` is selecting (`IsNearInteractorSelecting()`)

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -25,49 +28,67 @@ public class NearInteractionModeDetector : ProximityDetector
         /// colliders stopped being detected and update their buttons front plate and rounded
         /// rect if they are labeled as dynamic (based on proximity).
         /// </summary>
-        List<Collider> previouslyDetectedColliders = new List<Collider>();
+        private List<Collider> previouslyDetectedColliders = new List<Collider>();
 
         /// <inheritdoc />
         public override bool IsModeDetected()
```
keveleigh, line 34: Also curious about your thoughts on https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/611#discussion_r1489845921

> I'm not sure this is the best method to fire these events from, as it _could_ be called multiple times in a single frame. It also overloads the method a bit based on its simpler description: "Determine whether the detector has detected the appropriate conditions."

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
AMollis: I agree, we can avoid this issue by instead storing Interactables instead of colliders.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
AMollis: @ms-RistoRK ....If the foreach loop was changed to a for loop, and it iterates backwards, removal should avoid "skipping" colliders.

AMollis (changes requested):

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Nice catch! ... fixed in latest commit, resolving.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Actually, that's exactly why I had two collections at the very first version of this PR.  One collection would be used track which (Colllider, Interactor) stopped ProximityHovering on the interactable, the second collection was used to track which (Collider, Interactor) start ProximityHovering on the interactable.  After those collections were update in a frame, then in the next frame, the collections would be first checked for changes and then trigger the corresponding OnProximityHover{Enter|Exit} events.  I was precisely trying to prevent the traversal of a collection as it could still be changed.  Then I lost track of that approach as I implemented other feedback.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: > I'm also still curious about your thoughts on [#611 (comment)](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/611#discussion_r1489878783)
> 
> > Instead of sending multiple events to account for all `nearInteractors`, should we just send _one_ relaying info about either:
> > 
> > 1. The `ProximityDetector` itself. We could potentially expose the list of `nearInteractors` as an `IReadOnlyList` or something if we think the interactables need that info
> > 2. Specifically the `nearInteractor`(s) doing the selecting. They likely aren't all selecting at once, but one of the cases for this logic is that _any_ `nearInteractor` is selecting (`IsNearInteractorSelecting()`)

At an earlier version of this PR my original approach was to filter the number of events risen by storing unique (Collider + Interactor) duples on the PressableButton side that were starting/stopping a ProximityHover events and only rise any such event when the corresponding collections would add or remove a unique duple.  Such approach is lost in translation at this point after implementing other feedback.

ms-RistoRK: > I agree, we can avoid this issue by instead storing Interactables instead of colliders.

Could you please elaborate more on why Interactables will prevent the issue?

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: > Also curious about your thoughts on [#611 (comment)](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/611#discussion_r1489845921)
> 
> > I'm not sure this is the best method to fire these events from, as it _could_ be called multiple times in a single frame. It also overloads the method a bit based on its simpler description: "Determine whether the detector has detected the appropriate conditions."

At an earlier version of this PR my original approach was to filter the number of events risen by storing unique (Collider + Interactor) duples on the PressableButton side that were starting/stopping a ProximityHover events and only rise any such event when the corresponding collections would add or remove a unique duple.  Such approach is lost in translation at this point after implementing other feedback.  Granted, there were multiple calls to attempt to register a new (Collider + Interactor) duple but that would be prevented if the collection already had one such instance.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Switching from storing Interactor to storing Interactables in commit https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/e2f21e5a5f09660f1e9db8569e5a2584692067d7

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
AMollis: I think you are still missing colliders because you are not traversing backwards on `previouslyDetectedColliders`

`for (int i = 0; i < previouslyDetectedColliders.Count; i++)`

should be

`for (int i = previouslyDetectedColliders.Count; i >= 0; i--)`

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,84 @@ namespace MixedReality.Toolkit.Input
...
-                        IXRProximityInteractable nearInteractionMode = collider.GetComponent<IXRProximityInteractable>();
-                        if (nearInteractionMode != null)
-                        {
-                            foreach (XRBaseInteractor xrBaseInteractor in nearInteractors)
-                            {
-                                collider.GetComponent<IXRProximityInteractable>().OnProximityEntered(new ProximityEnteredEventArgs(this, collider, xrBaseInteractor));
-                            }
-                        }
-                        previouslyDetectedColliders.Add(collider);
+                        xrProximityInteractable.OnProximityEntered(new ProximityEnteredEventArgs(collider, xrBaseInteractable));
```
AMollis, line 81: You are still potentially calling `OnProximityEntered` twice on an interactable. if that interactable has multiple colliders.

I still think you should be storing `XRBaseInteractable` in `previouslyDetectedColliders`, instead of colliders.   I think this would help you prevent multiple calls to `OnProximityEntered`

AMollis (changes requested):

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Implemented in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/b4533fb2ae5a07724adf0bad32700d1ef82ea469

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Actually, we need to store both XRBaseInteractable and Colliders because such combination uniquely identifies the hover-proximity event for individual buttons, otherwise some buttons won't reset their state to non-hover-proximity and would only react to Hover events (I tested it).  We had the Interactable+Collider earlier but we lost while implementing feedback.  I have refactored the code to retake such approach so that button frontplate rawimages react to hover+proximity events as it was originally behaving.  The refactoring is implemented in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/2de988906343015208b3b1a5f6069fa09347d3ea resolving this for another pass on the PR.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,101 @@ namespace MixedReality.Toolkit.Input
...
 
         /// <summary>
-        /// Used to keep track of the previously detected colliders so that we can know which
-        /// colliders stopped being detected and update their buttons front plate and rounded
-        /// rect if they are labeled as dynamic (based on proximity).
+        /// Used to keep track of the previously detected colliders+interactrables duples (CID) so that we can know which
+        /// CID stopped being detected and update their buttons front plate RawImage.
         /// </summary>
-        List<Collider> previouslyDetectedColliders = new List<Collider>();
+        private HashSet<(Collider, IXRProximityInteractable)> previouslyDetectedColliderInteractableDuples = new HashSet<(Collider, IXRProximityInteractable)>();
```
AMollis, line 30: Why are you storing both Collider and IXRProximityInteractable?

You should be able to do what you need by only storing IXRProximityInteractable.   Perhaps we should chat.

AMollis (changes requested):

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
AMollis: I still don't think you need to store the collider at all...just the interactable....NOT the Interactor.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: That's a remnant from the initial approach, I was under the impression that it was needed to uniquely identify which UIElement was receiving the hoverProximity event and when trying to access both the FrontPlate RawImage and the Canvas' CanvasRoundedRect, now I realize it is indeed not needed for just modifying the FrontPlate's RawImage.  I've removed it in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/460c96e37cfc4d3ad2b7bfb54cdc14d4e9ed733d

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Correct.  That's a remnant from the initial approach, I was under the (wrong) impression that it was needed to uniquely identify which UIElement was receiving the hoverProximity event and when trying to access both the FrontPlate RawImage and the Canvas' CanvasRoundedRect, now I realize it is indeed not needed for just modifying the FrontPlate's RawImage.  I've removed it in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/460c96e37cfc4d3ad2b7bfb54cdc14d4e9ed733d

org.mixedrealitytoolkit.input/InteractionModes/IXRProximityInteractable.cs
```diff
@@ -18,15 +18,13 @@ public interface IXRProximityInteractable
         /// <summary>
         /// Registers the duple Collider + XRBaseInteractor as triggering proximity.
         /// </summary>
-        /// <param name="collider">Collider triggering proximity.</param>
-        /// <param name="xrBaseInteractor">Interactor triggering proximity.</param>
-        void OnProximityEntered(ProximityEnteredEventArgs args);
+        /// <param name="args">ProximityHoverEnteredEventArgs that has the Collider + Interactable duple triggering the proximity-hover event.</param>
```
keveleigh, line 21: Is this comment still true? I don't think it has the `Collider` anymore

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
```diff
@@ -640,38 +640,37 @@ private float IntersectQuad(Vector3 start, Vector3 end, float depth)
...
-        /// <param name="xrBaseInteractor">Interactor triggering proximity.</param>
-        public void OnProximityEntered(ProximityEnteredEventArgs proximityEnteredEventArgs)
+        /// <param name="xrBaseInteractable">Interactable triggering proximity.</param>
+        public void OnProximityEntered(ProximityHoverEnteredEventArgs proximityEnteredEventArgs)
         {
-            if (proximityEnteredEventArgs.collider != null &&
-                activeCollidersWithInteractor.Add((proximityEnteredEventArgs.collider, proximityEnteredEventArgs.interactor)) &&
-                activeCollidersWithInteractor.Count >= 1)
+            if (activeInteractables.Add(proximityEnteredEventArgs.Interactable as XRBaseInteractable) &&
+                activeInteractables.Count >= 1)
```
keveleigh, line 649: nit: do we need to check the `Count` if the `Add` was successful? Feels like that would guarantee a count of at least 1

org.mixedrealitytoolkit.input/InteractionModes/IXRProximityInteractable.cs
ms-RistoRK: Nope, not anymore.  I missed that when updating comments in a previous commit.  Thanks for catching that and letting me know.  Fixed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/29a7dee6124468197e8ac62359687d173eeb50c4

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
ms-RistoRK: Not really.  It is not needed, that's a remnant of then activeInteractables was a List<>.  I have update the code in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/d43041173f5cf1176e155c51ff57d1b618a7d29d resolving.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,99 @@ namespace MixedReality.Toolkit.Input
...
+            return result;
+        }
+
+        /// <summary>
+        /// Calls OnProximityExited for all interactables that were previously ProximityHover detected but are no longer detected.
+        /// </summary>
+        private void UpdateProximityExited()
+        {
+            List<IXRProximityInteractable> currentlyDetectedInteractable = GetCurrentlyDetectedColliderInteractablesDuples();
+            List<IXRProximityInteractable> noLongerDetectedInteractables = new();
```
AMollis, line 50: we should not be allocating a new list every frame.   Allocations have performance impact when done every frame.

If you want to do this "copy to list" pattern.   Make the list a class member, and clear it every frame instead of allocating.

Why did the reverse iteration not work?

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,99 @@ namespace MixedReality.Toolkit.Input
...
 
         /// <summary>
-        /// Used to keep track of the previously detected colliders so that we can know which
-        /// colliders stopped being detected and update their buttons front plate and rounded
-        /// rect if they are labeled as dynamic (based on proximity).
+        /// Used to keep track of the previously detected interactrables so that we can know which
+        /// interactable stopped being detected and update their buttons front plate RawImage.
         /// </summary>
-        List<Collider> previouslyDetectedColliders = new List<Collider>();
+        private HashSet<IXRProximityInteractable> previouslyDetectedColliderInteractableDuples = new();
```
AMollis, line 30: No longer a "Duple"

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,99 @@ namespace MixedReality.Toolkit.Input
...
+            }
+        }
+
+        /// <summary>
+        /// Returns a hashset of all unique interactables that are currently detected.
+        /// </summary>
+        /// <returns>Hashset with unique interactables currently detected</returns>
+        private List<IXRProximityInteractable> GetCurrentlyDetectedColliderInteractablesDuples()
+        {
+            List<IXRProximityInteractable> result = new();
```
AMollis, line 88: Remove "per frame" allocation.   Instead have class member list that you only allocate once, and clear() every frame.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,99 @@ namespace MixedReality.Toolkit.Input
...
+            }
+            return result;
+        }
+
+        /// <summary>
+        /// Calls OnProximityExited for all interactables that were previously ProximityHover detected but are no longer detected.
+        /// </summary>
+        private void UpdateProximityExited()
+        {
+            List<IXRProximityInteractable> currentlyDetectedInteractable = GetCurrentlyDetectedColliderInteractablesDuples();
```
AMollis, line 49: You're calling GetCurrentlyDetectedColliderInteractablesDuples() twice per frame.  Call once, and share the result with UpdateProximityExited() and UpdateProximityEntered()

AMollis (changes requested):

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: oops, I missed that.  Fixed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/a158aa6824f0921963dd143379e87db115469969 resolving this.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/7ee0fe33c7946426a3a643e4c266e4defd1fc3e2 resolving

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/7ee0fe33c7946426a3a643e4c266e4defd1fc3e2 resolving

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: I missed that.  Moved the datastructures to class-level variables in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/7ee0fe33c7946426a3a643e4c266e4defd1fc3e2
Re: Why did the reverse iteration not work? ... I think it was working, however, since there was feedback from another reviewer that there were missing colliders from processing I decided to reduce the number moving parts in the refactoring and go the safe route.  Would you like me to retry with a reverse iteration?
In the meantime, resolving this conversation (please feel free to open a new conversation if you would like me to tackle the reverse iteration approach) since the issue of allocating a new list every frame is addressed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/7ee0fe33c7946426a3a643e4c266e4defd1fc3e2

AMollis (approved):
@keveleigh can you also take a look at this?

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,106 @@ namespace MixedReality.Toolkit.Input
...
     public class NearInteractionModeDetector : ProximityDetector
     {
+        /// <summary>
+        /// The set of near interactables that belongs to near interaction
+        /// </summary>
         [SerializeField]
-        [Tooltip("The set of near interactors that belongs to near interaction")]
-        private List<XRBaseInteractor> nearInteractors;
+        [Tooltip("The set of near interactables that belongs to near interaction")]
+        private List<XRBaseInteractable> nearInteractables;
```
keveleigh, line 24: Please let me know if I missed a comment thread here for context, but I'm not sure why this was changed from interactors to interactables. This would be a breaking change, and also one that makes this component very difficult to keep up to date.

Previously, while storing a serialized list of interactors, it's easy to ensure all near interactors for each hand are represented here, because there aren't many (our current default is two per-hand):

image

With this change, does that mean I need to serialize every single near interactable in every single scene I might be using? That feels quickly unwieldy. Where is this currently being filled?

It also prevents us from keeping the MRTK XR Rig prefab in a usable state out-of-the-box, as a developer will need to assign every interactable in the Inspector before use.

I strongly feel this should be reverted.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,106 @@ namespace MixedReality.Toolkit.Input
...
+        /// Indicates whether any near interactable has interactors selecting.
+        /// </summary>
+        /// <returns>True if an interactable has interactors selecting, false otherwise.</returns>
         private bool IsNearInteractorSelecting()
         {
-            foreach (XRBaseInteractor nearInteractor in nearInteractors)
+            foreach (XRBaseInteractable nearInteractable in nearInteractables)
             {
-                if (nearInteractor.hasSelection)
+                if (nearInteractable.interactorsSelecting.Count > 0)
```
keveleigh, line 118: There's no guarantee that the interactor selecting this interactable has to be a near interactor. Many interactors can be selected both by near and far interaction, even simultaneously.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,106 @@ namespace MixedReality.Toolkit.Input
...
+                if (!currentlyDetectedInteractables.Contains(previouslyDetectedInteractable))
+                {
+                    noLongerDetectedInteractables.Add(previouslyDetectedInteractable);
+                }
+            }
+
+            foreach (IXRProximityInteractable noLongerDetectedInteractable in noLongerDetectedInteractables)
+            {
+                noLongerDetectedInteractable.OnProximityExited(new ProximityHoverExitedEventArgs(noLongerDetectedInteractable));
+                currentlyDetectedInteractables.Remove(noLongerDetectedInteractable);
```
keveleigh, line 73: Do we need this `Remove` call? We only add to `noLongerDetectedInteractables` if `currentlyDetectedInteractables` doesn't contain the interactable (checked on line 64)

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,106 @@ namespace MixedReality.Toolkit.Input
...
+            }
+        }
+
+        /// <summary>
+        /// Call OnProximityEntered for all interactables that are currently detected but were not detected previously.
+        /// </summary>
+        private void UpdateProximityEntered()
+        {
+            foreach (IXRProximityInteractable currentlyDetectedInteracable in currentlyDetectedInteractables)
```
keveleigh, line 82: nit: typo in `currentlyDetectedInteracable` (missing the final `t` in `Interactable`)

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,63 +16,106 @@ namespace MixedReality.Toolkit.Input
...
+            {
+                if (previouslyDetectedInteractables.Add(currentlyDetectedInteracable))
```
keveleigh, line 84: Do we ever need to clear or remove from `previouslyDetectedInteractables`? On a quick readthrough, it feels like we'll only send `OnProximityEntered` the first time an interactable is hovered because it'll stay in this list forever.

keveleigh: Oh, is the line above where I asked about `currentlyDetectedInteractables.Remove` supposed to be `previouslyDetectedInteractables.Remove`?

keveleigh (changes requested):

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
AMollis: Ack I missed this!   Yesh we should not be changing that..

AMollis (changes requested):
please make the changes suggested by @keveleigh

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
AMollis: I suspect this was a Find and Replace error

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Fixed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/996c8069b8dddc09d3aeaf7977f2114842cfd025, resolving.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: It is a small optimization.  After UpdateProximityExited() is called UpdateProximityEntered() is called and it has a cycle that traverses currentlyDetectedInteractables, by removing it in UpdateProximityExited() we avoid processing it in UpdatePrxoimityEntered().  I can remove it though, Would you like me to remove it?

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Yes, you are correct, I missed that.  Fixed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/72b8f65ef00615ff47b49f537b66da5c33bc1e23, resolving this.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Reverted in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/ee2f619a385491f73d7303e544ff944704e0042f, resolving.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Reverted in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/ee2f619a385491f73d7303e544ff944704e0042f, resolving.

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
```diff
@@ -640,38 +640,36 @@ private float IntersectQuad(Vector3 start, Vector3 end, float depth)
...
-        /// <param name="collider">Collider triggering proximity.</param>
-        /// <param name="xrBaseInteractor">Interactor triggering proximity.</param>
-        public void OnProximityEntered(ProximityEnteredEventArgs proximityEnteredEventArgs)
+        /// <param name="xrBaseInteractable">Interactable triggering proximity.</param>
+        public void OnProximityEntered(ProximityHoverEnteredEventArgs proximityEnteredEventArgs)
         {
-            if (proximityEnteredEventArgs.collider != null &&
-                activeCollidersWithInteractor.Add((proximityEnteredEventArgs.collider, proximityEnteredEventArgs.interactor)) &&
-                activeCollidersWithInteractor.Count >= 1)
+            if (activeInteractables.Add(proximityEnteredEventArgs.Interactable as XRBaseInteractable))
```
AMollis, line 648: The interactable here will always be the button itself.   The args should be the interactor (or the ProximityDetector) that is proximity hovering over the button.

Why was this change made?

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,58 +16,101 @@ namespace MixedReality.Toolkit.Input
...
-                        if (nearInteractionMode != null)
-                        {
-                            foreach (XRBaseInteractor xrBaseInteractor in nearInteractors)
-                            {
-                                previouslyDetectedCollider.GetComponent<IXRProximityInteractable>().OnProximityExited(new ProximityExitedEventArgs(this, previouslyDetectedCollider, xrBaseInteractor));
-                            }
-                        }
-                        previouslyDetectedColliders.Remove(previouslyDetectedCollider);
-                    }
+                    currentlyDetectedInteractable.OnProximityEntered(new ProximityHoverEnteredEventArgs(currentlyDetectedInteractable));
```
AMollis, line 86: I just realized that you are passing the "interactable" into the args.   I don't see how that helps.  The interactable will always be the target object.   

The args should contain the thing doing the proximity hovering, the interactor or the ProximityDetector

AMollis (changes requested):

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
```diff
@@ -182,9 +182,9 @@ public SpaceMode DistanceSpaceMode
         private float WorldToLocalScale => transform.InverseTransformVector(transform.forward).magnitude;
 
         /// <summary>
-        /// Holds the duples Collider + Interactor that have triggered proximity.
+        /// Holds the Interactables that trigger proximity.
         /// </summary>
-        private HashSet<(Collider, XRBaseInteractor)> activeCollidersWithInteractor = new HashSet<(Collider, XRBaseInteractor)>();
+        private HashSet<XRBaseInteractable> activeInteractables = new HashSet<XRBaseInteractable>();
```
whebertML, line 187: How would other Interactables cause proximity with this Interactable?  It doesn't seem to make sense that PressableButton (an Interactable) would keep a set of other Interactables in proximity.  It made sense before to keep a set of colliders and associated Interactors that are in proximity.  What is this change addressing?

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
AMollis: `ProximityHoverEnteredEventArgs(this));`

ms-RistoRK: I was under the wrong impression that an Interactable would be enough to uniquely identify interaction and got confused with some previous feedback.  I reverted back to use Interactors in commit https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/21f16fc4597f04b28bbe910fcfd1c1bc6a21adf0
Thank you for the offline explanation on this issue btw.

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
ms-RistoRK: I got confused from a previous feedback and was under the wrong impression that and Interactable would be enough to uniquely identify the source of the event but I was wrong.  I reverted to using Interactors in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/21f16fc4597f04b28bbe910fcfd1c1bc6a21adf0
Thank you for the offline explanation on this issue by the way.

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
ms-RistoRK: Thank you for contributing to the conversation.  I got confused other feedback and also was under the wrong impression that an interactable would be enough to uniquely identify the source of the interaction and hence proposed this change.  I reverted to use interactors in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/21f16fc4597f04b28bbe910fcfd1c1bc6a21adf0 resolvin this.

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
```diff
@@ -182,9 +182,9 @@ public SpaceMode DistanceSpaceMode
         private float WorldToLocalScale => transform.InverseTransformVector(transform.forward).magnitude;
 
         /// <summary>
-        /// Holds the duples Collider + Interactor that have triggered proximity.
+        /// Holds the interactors that trigger proximity.
         /// </summary>
-        private HashSet<(Collider, XRBaseInteractor)> activeCollidersWithInteractor = new HashSet<(Collider, XRBaseInteractor)>();
+        private HashSet<NearInteractionModeDetector> activeInteractors = new HashSet<NearInteractionModeDetector>();
```
AMollis, line 187: Rename `activeInteractors` to `activeDectectors`

AMollis (changes requested):
Super small change

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
ms-RistoRK: Done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/673633711a758de353c35b7f90ab544115cf627b, resolving.

AMollis (dismissed):

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
```diff
@@ -640,38 +640,36 @@ private float IntersectQuad(Vector3 start, Vector3 end, float depth)
         }
 
         /// <summary>
-        /// Registers the duple Collider + XRBaseInteractor as triggering proximity.
+        /// Registers the interactable triggering proximity.
```
whebertML, line 643: "NearInteractionModeDetector" instead of "Interactable"?

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
```diff
@@ -640,38 +640,36 @@ private float IntersectQuad(Vector3 start, Vector3 end, float depth)
...
-                activeCollidersWithInteractor.Count >= 1)
+            if (activeDetectors.Add(proximityEnteredEventArgs.NearInteractionModeDetector))
             {
                 UpdateProximityHovered();
             }
         }
 
         /// <summary>
-        /// Unregisters the duple Collider + XRBaseInteractor as triggering proximity.
+        /// Unregisters the interactable triggering proximity.
```
whebertML, line 655: "NearInteractionModeDetector" instead of "Interactable"?

org.mixedrealitytoolkit.input/InteractionModes/IXRProximityInteractable.cs
```diff
@@ -16,17 +16,15 @@ namespace MixedReality.Toolkit.Input
     public interface IXRProximityInteractable
     {
         /// <summary>
-        /// Registers the duple Collider + XRBaseInteractor as triggering proximity.
+        /// Registers the interactable as triggering hover-proximity.
```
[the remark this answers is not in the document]
whebertML: Or perhaps just "Detector"?

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,58 +16,101 @@ namespace MixedReality.Toolkit.Input
...
+            {
+                if (!currentlyDetectedInteractables.Contains(previouslyDetectedInteractable))
+                {
+                    noLongerDetectedInteractables.Add(previouslyDetectedInteractable);
+                }
+            }
+
+            foreach (IXRProximityInteractable noLongerDetectedInteractable in noLongerDetectedInteractables)
+            {
+                noLongerDetectedInteractable.OnProximityExited(new ProximityHoverExitedEventArgs(this));
```
whebertML, line 72: Does this component protect against previous objects having been destroyed?  Is it possible that the noLongerDetectedInteractable is destroyed at this time, so calling OnProximityExited() will cause a null ref?

Perhaps we should check for Unity's overridden object null check before proceeding?
`if (noLongerDetectedInteractable != null) ...`

org.mixedrealitytoolkit.input/InteractionModes/BaseProximityEventArgs.cs
```diff
@@ -1,40 +1,27 @@
...
-using UnityEngine.XR.Interaction.Toolkit;
 
 namespace MixedReality.Toolkit.Input
 {
     /// <summary>
-    /// Event data associated with proximity events triggered by a Collider and an Interactor combo.
+    /// Event data associated with proximity events triggered by a NearInteractionModeDetector.
     /// </summary>
-    public abstract partial class BaseProximityEventArgs : EventArgs
+    public abstract class BaseProximityHoverEventArgs : EventArgs
```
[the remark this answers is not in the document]
ms-RistoRK: Darn it!, I lost that change in some local squash.  Thank you for catching that and letting me know.  Re-applied in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/d83c01a6a15b0c2b31148588e03a6b7626204f0a, resolving.

org.mixedrealitytoolkit.input/InteractionModes/ProximityExitedEventArgs.cs
```diff
@@ -1,29 +1,19 @@
+// Copyright (c) Mixed Reality Toolkit Contributors
+// Licensed under the BSD 3-Clause
+
 using UnityEngine;
-using UnityEngine.XR.Interaction.Toolkit;
 
 namespace MixedReality.Toolkit.Input
 {
-    public class ProximityExitedEventArgs : BaseProximityEventArgs
+    public class ProximityHoverExitedEventArgs : BaseProximityHoverEventArgs
```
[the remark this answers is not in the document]
ms-RistoRK: Darn it!, I lost that change in some local squash.  Thank you for catching that and letting me know.  Re-applied in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/d83c01a6a15b0c2b31148588e03a6b7626204f0a, resolving.

org.mixedrealitytoolkit.input/InteractionModes/ProximityEnteredEventArgs.cs
```diff
@@ -1,29 +1,19 @@
+// Copyright (c) Mixed Reality Toolkit Contributors
+// Licensed under the BSD 3-Clause
+
 using UnityEngine;
-using UnityEngine.XR.Interaction.Toolkit;
 
 namespace MixedReality.Toolkit.Input
 {
-    public class ProximityEnteredEventArgs : BaseProximityEventArgs
+    public class ProximityHoverEnteredEventArgs : BaseProximityHoverEventArgs
```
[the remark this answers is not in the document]
ms-RistoRK: Darn it!, I lost that change in some local squash.  Thank you for catching that and letting me know.  Re-applied in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/d83c01a6a15b0c2b31148588e03a6b7626204f0a, resolving.

org.mixedrealitytoolkit.input/InteractionModes/IXRProximityInteractable.cs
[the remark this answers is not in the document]
ms-RistoRK: Fixed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/655e552cbcb02e2bdf86e6af4a0a52bebf58f033, resolving.

org.mixedrealitytoolkit.input/InteractionModes/IXRProximityInteractable.cs
```diff
@@ -16,17 +16,15 @@ namespace MixedReality.Toolkit.Input
...
         /// </summary>
-        /// <param name="collider">Collider triggering proximity.</param>
-        /// <param name="xrBaseInteractor">Interactor triggering proximity.</param>
-        void OnProximityEntered(ProximityEnteredEventArgs args);
+        /// <param name="args">ProximityHoverEnteredEventArgs that has the Interactable triggering the proximity-hover event.</param>
```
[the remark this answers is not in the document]
ms-RistoRK: Went with 'Detector' in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/655e552cbcb02e2bdf86e6af4a0a52bebf58f033, resolving.

org.mixedrealitytoolkit.input/InteractionModes/IXRProximityInteractable.cs
```diff
@@ -16,17 +16,15 @@ namespace MixedReality.Toolkit.Input
...
+        void OnProximityEntered(ProximityHoverEnteredEventArgs args);
 
         /// <summary>
-        /// Unregisters the duple Collider + XRBaseInteractor as triggering proximity.
+        /// Unregisters the interactable as triggering hover-proximity.
```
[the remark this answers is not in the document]
ms-RistoRK: Went with 'Detector' in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/655e552cbcb02e2bdf86e6af4a0a52bebf58f033, resolving.

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
ms-RistoRK: Went with 'Detector' in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/655e552cbcb02e2bdf86e6af4a0a52bebf58f033, resolving.

org.mixedrealitytoolkit.input/InteractionModes/IXRProximityInteractable.cs
```diff
@@ -16,17 +16,15 @@ namespace MixedReality.Toolkit.Input
...
         /// </summary>
-        /// <param name="collider">Collider that in combination with the interactor was triggering proximity.</param>
-        /// <param name="xrBaseInteractor">Interactor that in combination with the collider was triggering proximity.</param>
-        void OnProximityExited(ProximityExitedEventArgs args);
+        /// <param name="args">ProximityHoverExitedEventArgs that has the Interactable triggering the proximity-hover event.</param>
```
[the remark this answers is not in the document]
ms-RistoRK: Went with 'Detector' in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/655e552cbcb02e2bdf86e6af4a0a52bebf58f033, resolving.

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
ms-RistoRK: Went with 'Detector' in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/655e552cbcb02e2bdf86e6af4a0a52bebf58f033, resolving.

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
ms-RistoRK: Add null check in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/c97fcc6a878f218626f5099c763284ad6118699d, resolving.

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
whebertML, line 643: Detector instead of interactable.  But preferable to just do `/// <inheritdoc/>`

ms-RistoRK: Updated in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/b12b1cea8efa1f7b52f98b647c5bf697efe4e844, resolving.

org.mixedrealitytoolkit.uxcore/Button/PressableButton.cs
```diff
@@ -640,38 +640,36 @@ private float IntersectQuad(Vector3 start, Vector3 end, float depth)
...
```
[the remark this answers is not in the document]
ms-RistoRK: https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/659/commits/b12b1cea8efa1f7b52f98b647c5bf697efe4e844

whebertML (approved):

AMollis (approved):

AMollis:
@keveleigh would be good to get your sign-off as well

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
```diff
@@ -16,58 +16,104 @@ namespace MixedReality.Toolkit.Input
...
-        List<Collider> previouslyDetectedColliders = new List<Collider>();
+        private readonly List<IXRProximityInteractable> currentlyDetectedInteractables = new();
+
+        /// <summary>
+        /// Stores the interactables that are no longer detected (updated every frame).
+        /// </summary>
+        private readonly List<IXRProximityInteractable> noLongerDetectedInteractables = new();
 
         /// <inheritdoc />
         public override bool IsModeDetected()
```
keveleigh, line 43: I still have some worry that this method is _not_ guaranteed to only be called once per-frame, but, in practice, each detector only has one controller assigned, so it _should_ only be once. We should keep an eye on that changing, though.

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/95db47bc70e9d1f0393a19534460b2c682f5c0c4/org.mixedrealitytoolkit.input/InteractionModes/InteractionModeManager.cs#L412-L418

keveleigh (approved):

org.mixedrealitytoolkit.input/InteractionModes/NearInteractionModeDetector.cs
keveleigh: This is also a `MonoBehaviour`. We _could_ run this code in `Update()` instead, to ensure once-per-frame.

ms-RistoRK: Thank you for the additional information.  We'll keep an eye on the controller-detector 1:1 relation changing.  In the meantime, resolving this conversation to be able to merge.

Issue #652, opened 2024-02-21
[BUG] Slider (non-canvas) wrong scaling


Daniel4144:
## Describe the bug

The slider components do not scale correctly (NonCanvasSliderBase, Slider_168x40mm_Base), see screenshot below. The line does not fit with scales other than 1.

## Screenshots

![grafik](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/assets/49939834/3901f910-e4b4-4a32-a7da-cb5842cb6fa8)
Scale = 2

## Your setup (please complete the following information)

- Unity Version 2022.3.12f1
- MRTK UX  v3.1.1

AMollis:
@Daniel4144 , is the problem that the backplate does not scale with the slider bar?

Daniel4144:
No, it is only the line of the slider that does not scale correctly. I looked into it and found that its local scale is applied depending on the world positions of the Start and End transforms.

Transforming the SliderTrackDirection vector to local coordinates fixes the issue:
  ```csharp
//SliderVisuals.cs line 89
Vector3 localSliderTrackVector = trackArea.parent.InverseTransformVector(SliderState.SliderTrackDirection); //does trackArea always have a parent?
trackArea.localScale = new Vector3(localSliderTrackVector.magnitude, trackArea.localScale.y, trackArea.localScale.z);
```

AMollis:
@Daniel4144 , if you have a potential fix, feel free to make a Pull Request, and we will review.    This might be quicker than waiting for one of the maintainers to fix.

PR #664, opened 2024-02-29
fix for erroneous scaling of slider visual

Daniel4144:
fix for #652

### DIFF
--- a/org.mixedrealitytoolkit.uxcore/Slider/Visuals/SliderVisuals.cs
+++ b/org.mixedrealitytoolkit.uxcore/Slider/Visuals/SliderVisuals.cs
@@ -86,7 +86,8 @@ namespace MixedReality.Toolkit.UX
             trackArea.transform.localRotation = Quaternion.Euler(new Vector3(0, 0, angle));
             handle.transform.localRotation = Quaternion.Euler(new Vector3(0, 0, angle));
 
-            trackArea.localScale = new Vector3(SliderState.SliderTrackDirection.magnitude, trackArea.localScale.y, trackArea.localScale.z);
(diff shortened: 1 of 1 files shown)


org.mixedrealitytoolkit.uxcore/Slider/Visuals/SliderVisuals.cs
```diff
@@ -86,7 +86,8 @@ private void Update()
             trackArea.transform.localRotation = Quaternion.Euler(new Vector3(0, 0, angle));
             handle.transform.localRotation = Quaternion.Euler(new Vector3(0, 0, angle));
 
-            trackArea.localScale = new Vector3(SliderState.SliderTrackDirection.magnitude, trackArea.localScale.y, trackArea.localScale.z);
+            Vector3 localSliderTrackVector = trackArea.parent.InverseTransformVector(SliderState.SliderTrackDirection); //does trackArea always have a parent?
```
keveleigh, line 89: > //does trackArea always have a parent?

Good question! It's probably worth null-checking and only taking this path if it's non-null. Otherwise, follow the existing logic

Daniel4144: I've added the null check

keveleigh (approved):

org.mixedrealitytoolkit.uxcore/Slider/Visuals/SliderVisuals.cs
keveleigh: Seems good to me, thanks!

PR #706, opened 2024-03-26
Made bounds control more overridable

HossamBasion:
Made four functions in boundscontrol.cs protected virtual.

This is necessary for adding custom translation, rotation and scaling logic for bound controls behavior.

### DIFF
--- a/com.microsoft.mrtk.spatialmanipulation/BoundsControl/BoundsControl.cs
+++ b/com.microsoft.mrtk.spatialmanipulation/BoundsControl/BoundsControl.cs
@@ -530,7 +530,7 @@ namespace Microsoft.MixedReality.Toolkit.SpatialManipulation
             }
         }
 
-        private void Update()
+        protected virtual void Update()
         {
             // If we need to recompute bounds (usually because we found a
             // UGUI element), make sure we've waited enough frames since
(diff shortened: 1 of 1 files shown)


com.microsoft.mrtk.spatialmanipulation/BoundsControl/BoundsControl.cs
```diff
@@ -734,7 +734,7 @@ internal void OnHandleSelectEntered(BoundsHandleInteractable handle, SelectEnter
         private static readonly ProfilerMarker TransformTargetPerfMarker =
             new ProfilerMarker("[MRTK] BoundsControl.TransformTarget");
 
-        private void TransformTarget()
+        protected virtual void TransformTarget()
```
AMollis, line 737: I'm a little worried about exposing this function.  Can you please explain what behavior you're trying to override.   Perhaps there's a cleaner way to do this instead of exposing this function.

AMollis (changes requested):
Please open a a bug/feature request for us to track this PR.

Also, please update the package's changelog.md file, and check the version in package.json.  I believe the package's minor version needs to be updated.  Please reference to the current "release" branch for the last good version, the minor version should be one more (https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/releases/2024-03-20/org.mixedrealitytoolkit.spatialmanipulation/package.json)

com.microsoft.mrtk.spatialmanipulation/BoundsControl/BoundsControl.cs
HossamBasion: Hi @AMollis,

We want to implement custom rotation, scaling, translation methods in different use cases

shaynie:
@HossamBasion, please re-target your pull request to target the main branch

HossamBasion:
@shaynie Sounds good.

HossamBasion:
Recreating and retargetting to latest main branch

PR #719, opened 2024-04-07
Updated ConstantViewSize solver to retain the initial scale and aspect ratio (Fixes #306)

### LINKED ISSUE #306: ConstantViewSize solver changes the scale of the object to a ratio of (1:1:1) on three axes
### Original issue opened by:
@KritiaSthovania @KritiaSthovania
___

## Describe the bug

The _ConstantViewSize_ solver automatically changes the scale of the object to a ratio of (1:1:1) on three axes, as it directly computes the _GoalScale_ using _Vector3.one*scale_. However, there can be problems if the original scale of the object/prefab is not (1,1,1).

## To reproduce

1. Choose an object in the scene.
2. Change its scale to (1,2,1) in the inspector.
3. Add the _ConstantViewSize_ solver and _SolverHandler._
4. Start the scene. 
5. Seeing that the scale is modified to a ratio of (1:1:1).

## Expected behavior

The _ConstantViewSize_ solver can modify the scale of an object keeping its original ratio on three axes unchanged.

## Your setup

- Unity Version 2020.3.24f1
- MRTK Version 2.7.3

## Target platform

- HoloLens
- HoloLens 2

___
### ISSUE MIGRATION

   Issue migrated from: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/10396

BenediktTobias:
Fixes #306 

The ConstantViewSize solver now retains the original scale and aspect ratio as set by the base class Solver instead of using Vector3.one as the base scale.

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/Solvers/ConstantViewSize.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/Solvers/ConstantViewSize.cs
@@ -119,9 +119,9 @@ namespace MixedReality.Toolkit.SpatialManipulation
         /// Get the current scale progress on this <see cref="ConstantViewSize"/> object.
         /// </summary>
         /// <remarks>
-        /// This is a value between zero and one, representing progress between 
+        /// This is a value between zero and one, representing progress between
         /// <see cref="MinScale"/> and <see cref="MaxScale"/>. If current scale is less
-        /// than max,  then scaling is being applied. This value is subject to inaccuracies
+        /// than max, then scaling is being applied. This value is subject to inaccuracies
         /// due to smoothing, interpolation, and momentum.
         /// </remarks>
         public float CurrentScalePercent { get; private set; } = 1f;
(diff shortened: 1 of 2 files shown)


org.mixedrealitytoolkit.spatialmanipulation/Solvers/ConstantViewSize.cs
```diff
@@ -153,9 +153,17 @@ public float FovScale
...
 
+        private Vector3 originalScale;
         private float fovScalar = 1f;
         private float objectSize = 1f;
 
+        /// <inheritdoc/>
+        protected override void Awake()
+        {
+            base.Awake();
+            originalScale = GoalScale;
```
AMollis, line 164: I don't have enough context on the original intent here.

@keveleigh  or @david-c-kline ,  how do you feel about this change?

keveleigh: Seems good to me! Looks like an oversight in the original implementation, imo

keveleigh (approved):

PR #743, opened 2024-05-15
Upgrading to com.unity.xr.interaction.toolkit 3.0.1 - Step 2

ms-RistoRK:
Aiming for the goal of deprecating MRTK3 controllers, in this PR:
* Move ArticulatedHandController::PinchSelectReady property to GazePinchInteractor.
* Ignored tests that are failing because MRTK3 Gaze Controller is not fully migrated yet.
* Updated experimental MRTKXRIGaze prefab by:
  * Removing TrackedPoseDriver component to GazeInteractor component, MRTKXRIGaze prefab now looks like this:
image
  * Added actions to GazeInteractor InputConfiguration which is the new home for deprecated XRController actions as per [Slack channel with Unity](https://app.slack.com/client/E016WLPF0G6/C02GHQK9CP3), the experimental MRTKXRIGaze/GazeInteractor component now looks like this:
image

Manual testing done:  Manually tested in HL2 the experimental XRI prefabs in all MRTK3 main scenes via sideloaded app; all scenes work as expected, no functionality was impaired, no broken assets detected.


org.mixedrealitytoolkit.input/Tests/Runtime/BasicInputTests.cs
```diff
@@ -279,6 +279,8 @@ public IEnumerator StatefulInteractableSmoketest()
         /// Simple smoketest to ensure basic gaze-pinch selection functionality.
         /// </summary>
         [UnityTest]
+        [Ignore("Temporarily ignoring this test because MRTKGazeController is not fully migrated to XRI 3 yet.")]
```
AMollis, line 282: Can you open a bug to track reenabling these.  So we don't forget

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -28,12 +28,43 @@ public class GazePinchInteractor :
...
+            set
+            {
+                m_ControllerState = value;
+                m_CreateControllerState = false;
+            }
+        }
+
+        bool m_CreateControllerState = true;
+
+        void SetupControllerState()
```
AMollis, line 54: style....put functions down below, near other private functions.   And explicitly mark this as "private"

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -28,12 +28,43 @@ public class GazePinchInteractor :
         [Tooltip("The hand controller used to get the selection progress values")]
         private ArticulatedHandController handController;
 
+        #region Properties and methods from deprecated XRBaseController
+
+        XRControllerState m_ControllerState;
+        /// <summary>
+        /// The current state of the controller.
+        /// </summary>
+        public XRControllerState currentControllerState
```
AMollis, line 37: I'm pretty sure all controller types, including XRControllerState, were deprecated by Unity.  We shouldn't be adding "controllers state" to an interactor if deprecated.

AMollis: Also I don't understand the need to even add this state object as public.   Let's not add this, as I see no usage.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -28,12 +28,43 @@ public class GazePinchInteractor :
...
+
+        #endregion
+
         /// <summary>
         /// Is the hand ready to select? Typically, this
         /// represents whether the hand is in a pinching pose,
         /// within the FOV set by the aggregator config.
         /// </summary>
-        protected bool PinchReady => handController.PinchSelectReady;
+        protected bool PinchReady => (currentControllerState is ArticulatedHandControllerState handControllerState) && handControllerState.PinchSelectReady;
```
AMollis, line 67: Pinch ready is the only state needed here.   Let's just store the "bool" value, and update PinchReady in this interactor.

org.mixedrealitytoolkit.input/Controllers/ArticulatedHandController.cs
```diff
@@ -30,13 +30,6 @@ public class ArticulatedHandController : ActionBasedController
         /// <remarks>Expected to be XRNode.LeftHand or XRNode.RightHand.</remarks>
         public XRNode HandNode => handNode;
 
-        /// <summary>
-        /// Is the hand ready to select? Typically, this
-        /// represents whether the hand is in a pinching pose,
-        /// within the FOV set by the aggregator config.
-        /// </summary>
-        public bool PinchSelectReady => (currentControllerState is ArticulatedHandControllerState handControllerState) && handControllerState.PinchSelectReady;
```
AMollis, line 38: Do not remove functionality from our deprecated classes.  Leave as is and mark this class as deprecated.

org.mixedrealitytoolkit.input/Assets/Prefabs/Experimental-XRI3/SaveDevelopedPrefabsToExperimentalXRI3.ps1
```diff
@@ -0,0 +1,10 @@
+﻿Copy-Item '..\MRTK Gaze Controller.prefab' '.\MRTK XRI Gaze.prefab'
+Copy-Item '..\MRTK Interaction Manager.prefab' '.\MRTK XRI Interaction Manager.prefab'
```
AMollis, line 2: Why is this needed?  This copy is likely a onetime operation needed only be you.


Please don't include, so to avoid extra maintenance  cost.

org.mixedrealitytoolkit.input/Assets/Prefabs/Experimental-XRI3/MRTK XRI Gaze.prefab
AMollis, line 1: We need to rename this prefab as some point.   I missed it.


Don't include XRI on our prefab names.   The new Gaze prefab should have the same name as the old one, and the old one renames with "deprecated"

org.mixedrealitytoolkit.input/Assets/Prefabs/Experimental-XRI3/CopyNewPrefabsForDevelopment.ps1
```diff
@@ -0,0 +1,5 @@
+﻿Copy-Item '.\MRTK XRI Gaze.prefab' '..\MRTK Gaze Controller.prefab'
```
AMollis, line 1: Why is this needed? This copy is likely a onetime operation needed only be you.

Please don't include, so to avoid extra maintenance cost.

AMollis (changes requested):

org.mixedrealitytoolkit.input/Tests/Runtime/BasicInputTests.cs
ms-RistoRK: Sure, done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/744 , resolving.

org.mixedrealitytoolkit.input/Assets/Prefabs/Experimental-XRI3/CopyNewPrefabsForDevelopment.ps1
ms-RistoRK: Reverted with https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/ae962a4080fa69b164976601408ca163bb603ee0 , resolving.

org.mixedrealitytoolkit.input/Assets/Prefabs/Experimental-XRI3/SaveDevelopedPrefabsToExperimentalXRI3.ps1
ms-RistoRK: Reverted in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/ae962a4080fa69b164976601408ca163bb603ee0 , resolving.

org.mixedrealitytoolkit.input/Assets/Prefabs/Experimental-XRI3/MRTK XRI Gaze.prefab
ms-RistoRK: Done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/5d30b2aa0b219b20638a2df6aa6cdce33058de34 , resolving.

org.mixedrealitytoolkit.input/Controllers/ArticulatedHandController.cs
ms-RistoRK: Done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/a2d8ed684f96e6231f1bea0a283b8e0ed9b265ba , resolving.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: Done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/182e3c8900231046fe4cedb5ecdcb21797f6f946 , resolving.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: > I'm pretty sure all controller types, including XRControllerState, were deprecated by Unity. We shouldn't be adding "controllers state" to an interactor if deprecated.

I thought the same first but then I noticed that XRControllerState is not mentioned as part of the deprecation in Unity's changelog https://docs.unity3d.com/Packages/com.unity.xr.interaction.toolkit@3.0/changelog/CHANGELOG.html ... Also Visual Studio doesn't suggests it is deprecated as in:
image

> Also I don't understand the need to even add this state object as public. Let's not add this, as I see no usage.

I kept it public because that's how it is exposed in the deprecated XRController, my reasoning to keep it this way is that if we are going to remove XRController as part of XRI3 migration but some of our customers have code already depending on that property then helping them to upgrade their code to consume the replacement property(ies) would be easier if they have the same signature as they currently have in XRController.

@AMollis Given the above ^ explanation, Do you still want me to remove it or make it private?

ms-RistoRK: As per our offline discussion, I have avoided using the XRControllerState in GazePinchInteractor.  This is done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/07d41155987a3c194dbeb0c7404455a797418ada , resolving.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: Done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/07d41155987a3c194dbeb0c7404455a797418ada , resolving.

ms-RistoRK:
@keveleigh Would you like us to port the workaround for missing select actions on devices without interaction profiles [you recently touched](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/a02520f3bcad444c700a920d0adcaa02485ad3b1)? It would be lost once Controllers are removed for the XRI 3 migration.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -444,5 +452,64 @@ protected override void OnSelectExited(SelectExitEventArgs args)
...
+
+            if (XRSubsystemHelpers.HandsAggregator == null)
+            {
+                return;
+            }
+
+            // Determine the XRNode to use for the pinch data
+            bool gotPinchData;
+            XRNode xRNode;
+            if (handController != null) //For pre-XRI3 code.  Note: This if can be removed (leave the code in the 'else' section) when
```
AMollis, line 478: We don't need to handle Hand Controller here.  

The else case should work for both rigs.  Have you tried just relying on the stuff in the 'else'?

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -444,5 +452,64 @@ protected override void OnSelectExited(SelectExitEventArgs args)
...
+                xRNode = handController.HandNode;
+                if (xRNode != XRNode.LeftHand && xRNode != XRNode.RightHand)
+                {
+                    Debug.LogWarning($"HandController {handController.name} does not have HandNode set to neither XRNode.LeftHand nor XRNode.RightHand, defaulting to XRNode.RightHand.");
+                    xRNode = XRNode.RightHand;
+                }
+            }
+            else //For post-XRI3 code.  The XRNode is determined by the GazeInteractor Handedness if none is set then it defaults to XRNode.RightHand.
+            {
+                switch (handedness)
```
AMollis, line 490: make a helper extension for this.  And put in HandednessExtensions.cs

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -444,5 +452,64 @@ protected override void OnSelectExited(SelectExitEventArgs args)
...
+                        xRNode = XRNode.LeftHand;
+                        break;
+                    case InteractorHandedness.Right:
+                    default:
+                        xRNode = XRNode.RightHand;
+                        break;
+                }
+            }
+
+            gotPinchData = XRSubsystemHelpers.HandsAggregator.TryGetPinchProgress(xRNode, out bool isPinchReady, out bool isPinching, out float pinchAmount);
```
AMollis, line 502: XRSubsystemHelpers.HandsAggregator might be null...should protected against null case

AMollis (changes requested):

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: Already tested in line 470
image

ms-RistoRK: Add a console warning to the HandsAggregator null check in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/1dbab66bebc5577d7617f07e445b8dfae2091690 to facilitate debugging if HandsAggregator is null.  Resolving.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: I wasn't able to add to HandednessExtensions because GazePinchInteractor's handedness is of type InteractorHandedness and not of type Handedness.  So I created a new InteractorHandednessExtensions class and added it there.  Changes in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/6b3d24f5e3c85dd9d95dbc08721409bf828c1b96 , resolving.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: Indeed.  My bad, fixed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/338a5d70ac1bde41db85767a55abd3ee0994d21a , resolving.

org.mixedrealitytoolkit.core/Utilities/Extensions/InteractorHandednessExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Gets the <see cref="XRNode"/> representing the specified <see cref="InteractorHandedness"/>. If the <see cref="InteractorHandedness"/>
+        /// is other than InteractorHandedness.Left or InteractorHandedness.Right then it defaults to InteractorHandedness.Right.
+        /// </summary>
+        /// <param name="hand">The <see cref="InteractorHandedness"/> value for
+        /// which the <see cref="XRNode"/> is requested.</param>
+        /// <returns>
+        /// <see cref="XRNode"/> representing the specified <see cref="InteractorHandedness"/> with InteractorHandedness.Right as default.
+        /// </returns>
+        public static XRNode ToXRNodeWithRightHandDefault(this InteractorHandedness hand)
```
AMollis, line 23: This function name is a bit too verbose.  If you're wanting to add clarity here, perhaps have a default parameter.  How about this?

```cs
public static ToXRNode(this InteractorHandedness hand, XRNode defaultValue = XRNode.RightHand)
{
    switch (hand)
    {
        case InteractorHandedness.Left:
            return XRNode.LeftHand;
        case InteractorHandedness.Right:
            return XRNode.RightHand;
        default:
            return defaultValue;
    }
}
```

AMollis (changes requested):

org.mixedrealitytoolkit.core/Utilities/Extensions/InteractorHandednessExtensions.cs
ms-RistoRK: Love it, implemented in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/8938a7354c8ae853ff3df4eed5bb4a5c94312e3e , resolving.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -444,5 +452,38 @@ protected override void OnSelectExited(SelectExitEventArgs args)
...
+            {
+                Debug.LogWarning("XRSubsystemHelpers.HandsAggregator is null, pinch state won't update.");
+                return;
+            }
+
+            bool gotPinchData = XRSubsystemHelpers.HandsAggregator.TryGetPinchProgress(handedness.ToXRNode(),
+                out bool isPinchReady, out bool isPinching, out float pinchAmount);
+            if (!gotPinchData) //Try the other hand if the set hand does not have pinch data.
+            {
+                gotPinchData = XRSubsystemHelpers.HandsAggregator.TryGetPinchProgress(handedness.ToXRNode() == XRNode.LeftHand ? XRNode.RightHand : XRNode.LeftHand,
```
AMollis, line 479: you're converting the value twice.  store in `handedness.ToXRNode()` in local var.

AMollis (changes requested):
small change I missed earlier

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: Right, fixed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/743/commits/386494ad8a6d2591fb95d72335487b106ee6e339 , resolving.

AMollis (approved):

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -444,5 +452,39 @@ protected override void OnSelectExited(SelectExitEventArgs args)
...
+            if (XRSubsystemHelpers.HandsAggregator == null)
+            {
+                Debug.LogWarning("XRSubsystemHelpers.HandsAggregator is null, pinch state won't update.");
+                return;
+            }
+
+            var xrNode = handedness.ToXRNode();
+            bool gotPinchData = XRSubsystemHelpers.HandsAggregator.TryGetPinchProgress(xrNode,
+                out bool isPinchReady, out bool isPinching, out float pinchAmount);
+            if (!gotPinchData) //Try the other hand if the set hand does not have pinch data.
```
keveleigh, line 478: @ms-RistoRK I'm a little confused why we try the other hand. Aren't `GazePinchInteractor`s explicitly handed? So there should be a second one in the scene that will handle the other hand.

keveleigh:
> @keveleigh Would you like us to port the workaround for missing select actions on devices without interaction profiles [you recently touched](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/a02520f3bcad444c700a920d0adcaa02485ad3b1)? It would be lost once Controllers are removed for the XRI 3 migration.

I know I'm way behind here @ms-RistoRK, but did this happen and can you point me to it please?

PR #745, opened 2024-05-16
Updating MRTKRayInteractor to remove usage of deprecated controllers

AMollis:
_draft pr....not tested_

Updating MRTKRayInteractor to remove usage of deprecated controllers


AMollis:
@ms-RistoRK .... I took a stab at removing controller dependencies on MRTKRayInteractor.  Does require branching for old rig to work, but it's following the example in XRI3 and uses the `forceDeprecatedInput` field.

Issue #645, opened 2024-02-14
Plan out Unity 6.0 and XRI 3.0 support


AMollis:

shaynie:
@ms-RistoRK will be looking at this upgrade for the next couple sprints.
Expected to require major release bump
Talked to Unity:
Namespaces have changed
process is to upgrade to pre-release 2, THEN upgrade to release and then not running the migration tool.

bd-jimmy:
Would it make sense to upgrade the project to 2022 LTS and 2023 LTS first?  I see some slight upgrades for those versions.

shaynie:
Created feature/XRI3 branch on 4/17/24.

shaynie:
Risto is currently working on a new MRTK rig prefab without controller behaviors.
Old prefab will remain but deprecated

AMollis:
Before we ship this version....deprecated MRTK3 method/classes should be removed.

- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/531

shaynie:
Holding in discussion until the XRI 3 PR is committed

AMollis:
Updates the "details" of the issue, so goals are clearer

AMollis:
XRI3 branch still has some usages of deprecated XRI classes

**ObjectManipulator**

-  C:\dev\repos\mrtk3-atm\org.mixedrealitytoolkit.spatialmanipulation\ObjectManipulator\ObjectManipulator.cs(1065,17): warning CS0618: 'XRBaseInputInteractor.xrController' is obsolete: 'xrController has been deprecated in version 3.0.0.'
-  C:\dev\repos\mrtk3-atm\org.mixedrealitytoolkit.spatialmanipulation\ObjectManipulator\ObjectManipulator.cs(1065,54): warning CS0618: 'ActionBasedController' is obsolete: 'ActionBasedController has been deprecated in version 3.0.0. Its functionality has been distributed into different components.'

**RiggedHandMeshVisualizer**

- C:\dev\repos\mrtk3-atm\org.mixedrealitytoolkit.input\Visualizers\RiggedHandVisualizer\RiggedHandMeshVisualizer.cs(77,17): warning CS0618: 'XRBaseController' is obsolete: 'XRBaseController has been deprecated in version 3.0.0. Its functionality has been distributed into different components.'

**FlatScreenModeDetector**

[...]

Issue #531, opened 2023-10-25
Remove obsolete methods (with XRI 3.0 Updates)


david-c-kline:
It was originally intended that obsolete methods be removed prior to GA. Unfortunately, a few packages still have methods that should not be used.

NOTE: Removing these methods will require a _major_ version update for each impacted package.

> org.mixedrealitytoolkit.core\Utilities\Extensions\HandednessExtensions.cs:44:        [Obsolete("Use flags instead (e.g. Handedness.Right | Handedness.Left)")]
org.mixedrealitytoolkit.core\Utilities\Extensions\HandednessExtensions.cs:56:        [Obsolete("Use flags instead (e.g. Handedness.Right | Handedness.Left)")]
org.mixedrealitytoolkit.core\Utilities\Extensions\HandednessExtensions.cs:68:        [Obsolete("Use flags instead (e.g. Handedness.Right | Handedness.Left)")]
org.mixedrealitytoolkit.core\Utilities\HandsUtils.cs:22:        [Obsolete("Deprecated, please use XRSubsystemHelpers.HandsAggregator")]
org.mixedrealitytoolkit.core\Utilities\PlayspaceUtilities.cs:44:        [ObsoleteAttribute("For transforming poses into world space, use XROrigin.CameraFloorOffsetObject.transform.")]
org.mixedrealitytoolkit.core\Utilities\XRSubsystemHelpers.cs:55:        [Obsolete("GetAllRunningSubsystems now internally caches for you; this separate NonAlloc method is no longer necessary.")]
org.mixedrealitytoolkit.input\Controllers\ArticulatedHandController.cs:50:        [Obsolete("Deprecated, please use XRSubsystemHelpers.HandsAggregator instead.")]
org.mixedrealitytoolkit.input\Interactors\Ray\MRTKRayInteractor.cs:52:        [Obsolete("Deprecated, please use XRSubsystemHelpers.HandsAggregator instead.")]
org.mixedrealitytoolkit.input\Simulation\Devices\ControllerSimulationSettings.cs:61:        [Obsolete("This property is deprecated, use ToggleSecondaryHandshapes instead.")]
org.mixedrealitytoolkit.input\Subsystems\SyntheticHands\SyntheticHandsSubsystem.cs:76:        [Obsolete("Please use the GetNeutralHandshape(handNode) instead.")]
[...]

PR #746, opened 2024-05-17
Upgrading to com.unity.xr.interaction.toolkit 3.0.1 - Step 3

ms-RistoRK:
Aiming for the goal of deprecating MRTK3 controllers, in this PR:
  * Added TrackedPoseDriverWithFallback that will replace ActionBasedControllerWithFallbacks
  * Adding TrackedPoseDriverWithFallback component to Experimental-XRI3/MRTKGazeController/GazeInteractor prefab + linking Position, Rotation, and TrackingState actions and corresponding fallbacks.  The Experimental-XRI3/MRTKGazeController/GazeInteractor now looks like this:
image

 * Updated unity-tests so that the old functionality of ActionBasedControllerWithBallbacks functionality is still properly tested.

Note: Created an [issue for tracking the development of unity-tests for the new TrackedPoseDriverWithFallback once the new MRTKRig for XRI3 migration settles down](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/747).

Manual testing that this fallback functionality works was done with the following steps:
1 - Made a sideload install of HL2.
2 - Run the app and granted eye-tracking permissions.
3 - Opened a scene with multiple objects that when selected (via 'select' utterance) would be highlighted with a bounding box or other permanent visual until de-selected.
4 - With eye-gaze selected object with 'select' utterance.  This confirmed that eye-gaze input actions was providing data.
5 - Oriented HL2 toward an empty space.
6 - Closed my eyes and made 'select' utterance.
7 - Opened my eyes and confirmed nothing was selected.
8 - Oriented HL2 towards a selectable object.
9 - Closed my eyes and made 'select' utterance.
10 - Opened my eyes and confirmed that the object aligned in front of HL2 was selected.  This confirmed that when eye-gaze input actions did not provide data then the data from the fallback (head) input actions was used.


org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
AMollis, line 1: This is not the right folder for this.  Since it can be used for non-gaze scenario let's move it elsewhere.   

* [Tracking](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/tree/main/org.mixedrealitytoolkit.input/Tracking)
* [Utilities](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/tree/main/org.mixedrealitytoolkit.input/Utilities)

org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (!inputTrackingState.HasFlag(InputTrackingState.Position) && hasPositionAction && hasPositionFallbackAction)
+            {
+                (fallbackPositionAction, positionInput) = (positionInput, fallbackPositionAction);
+            }
+
+            // If no rotation data then swap the rotation action with the fallback rotation action if it exists
+            if (!inputTrackingState.HasFlag(InputTrackingState.Rotation) && hasRotationAction && hasRotationFallbackAction)
+            {
+                (fallbackRotationAction, rotationInput) = (rotationInput, fallbackRotationAction);
+            }
```
AMollis, line 94: I see what you're trying to do here...prevent failures in next update loop.  However, we shouldn't be changing public property values from under the consumers of this class.  Consumers can be caught off guard if these values change.

Also, with this there will be bugs.  If eye tracking is lost briefly, the app will be stuck in head tracking mode.  The app should revert back to eye tracking once available.   

Please revert this swapping logic

org.mixedrealitytoolkit.input/Tests/Runtime/FuzzyGazeInteractorTests.cs
```diff
@@ -160,6 +160,17 @@ public IEnumerator GazeFallbackWhenEyeGazeLosesTrackingTest()
         {
             // Confirm a FuzzyGazeInteractor is active in the scene
             FuzzyGazeInteractor fuzzyGazeInteractor = FindObjectUtility.FindFirstObjectByType<FuzzyGazeInteractor>();
+
+            #region TrackedPoseDriverWithFallback disabling
+            // Disable the TrackedPoseDriverWithFallback component that was added for XRI3 migration.  Disable it to ensure that
```
AMollis, line 165: We should be testing the new functionality too...where's this happening?

org.mixedrealitytoolkit.input/Assets/Prefabs/MRTK Gaze Controller.prefab
AMollis, line 1: Is this the old pre-XRI3 prefab or the new one?


We should NOT change pre-XRI3 prefabs

AMollis (changes requested):

org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
whebertML: Actually, scratch that first part.  I do see that the XRI3 rig is using TrackedPoseDriver at the top level as well, so a "WithFallback" might best go there.

But the second point, of starting to use more of the XRI3 rig directly and augment it as needed - is that in the XRI3 migration plan and at what point?

AMollis: @whebertML we aren't not planning to incorporate the XRI interactors.   This should be done in a separate release.   Per our discussion with the maintainers, we discussed the desire to use XRI3 interactors and follow the XRI3 rig example.  However, this is not in scope for this release.

whebertML: @AMollis   Thanks for clarification, I wasn't exactly sure when #67 was going to start happening.  Is there any documentation on the scope and goals of this first XRI3 migration release?  I'm not seeing many details in #645.  The related #531 is clear of course.

AMollis: @whebertML thanks for pointing out the lack detail in #645.   I've added our goals to this, please add comments to issue:

- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/645

whebertML: @AMollis Excellent, thanks!  Makes it much more clear!

org.mixedrealitytoolkit.input/Assets/Prefabs/MRTK Gaze Controller.prefab
ms-RistoRK: Right, reverted original MRTKGazeController and applied changes to Experimental-XRI3/MRTKGazeController prefab in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/746/commits/ae42cd46317bac59c81d20cfc8882f431946c053 , resolving.

org.mixedrealitytoolkit.input/Tests/Runtime/FuzzyGazeInteractorTests.cs
ms-RistoRK: Right now, in my dev desktop and in my dev HL2 to ensure functionality is as designed.  I swap the original prefabs with the experimental prefabs and run all Unity-tests as well make builds that I sideload to my HL2 and also the traditional stepXstep debugging with Visual Studio.  I haven't written specific unity-tests because the code is still evolving, although I did create a tracking issue for when the logic of this PR settles down:

image

org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
ms-RistoRK: I see your point.  I'm not sure I can just go back to the pre-XRI3 logic because we use the XRControllerState to replace the missing data with the fallback data and we are trying to stop using XRControllerState.  As in:
image
On the other hand, I did try intially to replicate the original logic but couldn't find a way to update the replace missing values directly in InputTrackingState nor the Actions themselves, for example:
image
Do you know a way to accomplish this without the XRControllerState?

ms-RistoRK: @AMollis:  I haven't been able to find a way to re-implement the original logic without an XRControllerState so I switched approaches: use InputActions with fallbacks and let the Input System get data from a priority list of bindings.  I provide more details in the Description of this PR.  What do you think of this alternate approach?

org.mixedrealitytoolkit.input/Assets/Input Actions/MRTK Default Input Actions.inputactions
```diff
@@ -1103,6 +1103,33 @@
                     "processors": "",
                     "interactions": "",
                     "initialStateCheck": true
+                },
+                {
```
AMollis, line 1107: @ms-RistoRK  this actually doesn't work 

The problem is that when eye's are calibrated the eyes are still available and unity tries to obtain data from them.  I needed to write the "fallback" script to check the tracking state of the eyes, and then fallback to the head pose.

Perhaps XRI3 is different. but sort of doubt it.

AMollis (changes requested):

org.mixedrealitytoolkit.input/Tests/Runtime/FuzzyGazeInteractorTests.cs
ms-RistoRK: As for the functionality of this PR (fallback action inputs) I am doing:
1 - Create a build with Experimental-XRI3/MRTKGazeController prefab.
2 - Sideload app to HL2
3 - Launch app
4 - Granted eye-tracking access to app
5 - Ran MRTK3 scenes
6 - While keeping the head steady used only eye-gaze + "Select" utterance to select different objects in the scene. This should succeed and confirm that the default eye-gaze input action is used.
7 - Uninstall app
8 - Re-install app from the same binaries
9 - Launch app
10 - Deny eye-tracking access to app
11 - Run MRTK3 scenes
12 - While keeping the head steady used only eye-gaze + "Select" utterance try to select different objects in the scene.  This should fail and confirm that eye-gaze input is not providing data.
13 - Align head to object that want to select, and used "Select" utterance.  This should succeed to confirm that the fallback head input action us used for input.

org.mixedrealitytoolkit.input/Assets/Input Actions/MRTK Default Input Actions.inputactions
ms-RistoRK: Got it!, thx for the clarification.  I'll retry the original approach then.  New approach reverted in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/746/commits/3ea4039f4cc9eea4de80226a30141eddce3d7308 , resolving.

org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
ms-RistoRK: Thank you for synching with me on this issue.  I've implemented the approach we discussed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/746/commits/23aec29f02753d453751f0feefe4130a3d6c8f2d and it works (steps on how I tested are in the description of this PR), resolving.

org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
```diff
@@ -0,0 +1,116 @@
...
+            // If no rotation data then use the data from the fallback action if it exists
+            if (!inputTrackingState.HasFlag(InputTrackingState.Rotation) && hasRotationFallbackAction)
+            {
+                neededToGetFallbackData = true;
+                rotation = fallbackRotationAction.action.ReadValue<Quaternion>();
+            }
+            else if (hasRotationAction)
+            {
+                rotation = rotationAction.ReadValue<Quaternion>(); //get a hold on the rotation data in case position data was obtained from fallback action
+            }
```
AMollis, line 107: I don't like handling the "default" case here.  We should allow the "base" functionality to work as is.

Take a look at the base.PerformUpdate() and observe how it is setting `transform.localPosition` and `transform.localRotation`.

Instead have the starting rotation equal `transform.localRotation`

`Quaternion rotation = transform.localRotation`

Then remove this `else` case.

org.mixedrealitytoolkit.input/Tests/Runtime/FuzzyGazeInteractorTests.cs
```diff
@@ -160,6 +160,17 @@ public IEnumerator GazeFallbackWhenEyeGazeLosesTrackingTest()
...
+            // the pre-XRI3 ActionBasedControllerWithFallbacks functionality is properly tested.
+            TrackedPoseDriverWithFallback trackedPoseDriverWithFallback = fuzzyGazeInteractor.GetComponent<TrackedPoseDriverWithFallback>();
+            if (trackedPoseDriverWithFallback != null)
+            {
+                trackedPoseDriverWithFallback.enabled = false;
```
AMollis, line 170: I still don't understand why this is needed, if we're not change the older prefab.   Since the older prefab is untouched, wouldn't the old tests continue to work?

org.mixedrealitytoolkit.input/Tests/Runtime/FuzzyGazeInteractorTests.cs
```diff
@@ -229,6 +240,17 @@ public IEnumerator GazeFallbackWhenEyeGazeIsDisabledTest()
         {
             // Confirm a FuzzyGazeInteractor is active in the scene
             FuzzyGazeInteractor fuzzyGazeInteractor = FindObjectUtility.FindFirstObjectByType<FuzzyGazeInteractor>();
+
+            #region TrackedPoseDriverWithFallback disabling
+            // Disable the TrackedPoseDriverWithFallback component that was added for XRI3 migration.  Disable it to ensure that
+            // the pre-XRI3 ActionBasedControllerWithFallbacks functionality is properly tested.
+            TrackedPoseDriverWithFallback trackedPoseDriverWithFallback = fuzzyGazeInteractor.GetComponent<TrackedPoseDriverWithFallback>();
+            if (trackedPoseDriverWithFallback != null)
+            {
```
AMollis, line 249: I still don't understand why this is needed, if we're not change the older prefab.   Since the older prefab is untouched, wouldn't the old tests continue to work?

org.mixedrealitytoolkit.input/Tests/Runtime/FuzzyGazeInteractorTests.cs
ms-RistoRK: You are right, I was under the wrong impression that the old prefabs would be changing.  Removed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/746/commits/5c757d880deef6cb6375b82340f815012616c1e9 , resolving.

org.mixedrealitytoolkit.input/Tests/Runtime/FuzzyGazeInteractorTests.cs
ms-RistoRK: You are right, I was under the wrong impression that the old prefabs would be changing.  Removed in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/746/commits/5c757d880deef6cb6375b82340f815012616c1e9 , resolving.

org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            if (neededToGetFallbackData) //because either position, rotation, or both data were obtained from fallback actions
+            {
+                SetLocalTransform(position, rotation);
```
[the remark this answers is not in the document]
ms-RistoRK: > I took a closer look at this `SetLocalTransform`, I actually think this is not going to work if the primary input is not tracked. See line 585 of TrackedPoseDriver.cs

That's why I added this condition in which if the primary input is not tracked then the InputTrackingState is ready from the FallbackTrackingStateAction if it exists:

image
 
> So I think we need to make our own `SetLocalTransformFromFallback` method, and call this instead:
> 
> ```cs
> protected virtual void SetLocalTransformFromFallback(Vector3 newPosition, Quaternion newRotation)
> {
>     var positionValid = this.IgnoreTrackingState || (m_CurrentFallbackTrackingState & TrackingStates.Position) != 0;
>     var rotationValid = this.IgnoreTrackingState || (m_CurrentFallbackTrackingState & TrackingStates.Rotation) != 0;
> 
> #if HAS_SET_LOCAL_POSITION_AND_ROTATION
>     if (this.TrackingType == TrackingType.RotationAndPosition && rotationValid && positionValid)
>     {
>         transform.SetLocalPositionAndRotation(newPosition, newRotation);
>         return;
>     }
> #endif
> 
>     if (rotationValid &&
>         (this.TrackingType == TrackingType.RotationAndPosition ||
>          this.TrackingType == TrackingType.RotationOnly))
>     {
>         transform.localRotation = newRotation;
>     }
> 
>     if (positionValid &&
>         (this.TrackingType == TrackingType.RotationAndPosition ||
>          this.TrackingType == TrackingType.PositionOnly))
>     {
>         transform.localPosition = newPosition;
>     }
> }
> ```

I think the condition I mentioned above does cover the case you are mentioning.  Could you please confirm whether it covers or not the case you mention?
Please let me know so that I can either revert or leave commit https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/746/commits/5ccb160ceddb922e1c03e7921e2010095954ec4e in which I implement your suggestion.

org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
```diff
@@ -0,0 +1,116 @@
...
+            bool neededToGetFallbackData = false;
+            Vector3 position = Vector3.zero;
+            Quaternion rotation = Quaternion.identity;
+
+            // If no position data then use the data from the fallback action if it exists
+            if (!inputTrackingState.HasFlag(InputTrackingState.Position) && hasPositionFallbackAction)
+            {
+                neededToGetFallbackData = true;
+                position = fallbackPositionAction.action.ReadValue<Vector3>();
+            }
```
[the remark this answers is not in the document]
ms-RistoRK: [donee](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/746/commits/758fcaff7754d6479e2a54e9636a20708e0ae828) , resolving

org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
ms-RistoRK: resolving https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/746/commits/758fcaff7754d6479e2a54e9636a20708e0ae828

org.mixedrealitytoolkit.input/Interactors/Gaze/TrackedPoseDriverWithFallback.cs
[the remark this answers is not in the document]
ms-RistoRK: done in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/746/commits/152f6436a5c3b1e38d0c4591bc3faa86aab00e90 resolving

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverWithFallback.cs
```diff
@@ -0,0 +1,168 @@
...
+            var hasPositionFallbackAction = fallbackPositionAction != null;
+
+            var hasRotationAction = rotationAction != null;
+            var hasRotationFallbackAction = fallbackRotationAction != null;
+
+            InputTrackingState inputTrackingState = (InputTrackingState)trackingStateInput.action.ReadValue<int>();
+            InputTrackingState fallbackInputTrackingState = InputTrackingState.None;
+
+            // If default InputTrackingState does not have position and rotation data, use fallback if it exists
+            if (!inputTrackingState.HasFlag(InputTrackingState.Position) && !inputTrackingState.HasFlag(InputTrackingState.Rotation) && FallbackTrackingStateAction.action != null)
```
AMollis, line 106: I don't think you need lines 106 -109.

Why are you overriding `inputTrackingState` with `FallbackTrackingStateAction`?

If inputTrackingState starts out with no valid position or rotation, then we want to use the  fallback.  However, if you're setting `inputTrackingState` to `FallbackTrackingStateAction`, then lines 123 - 124 and 130-131 will never execute.

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverWithFallback.cs
```diff
@@ -0,0 +1,168 @@
...
+    [DisallowMultipleComponent]
+    [AddComponentMenu("MRTK/Input/Tracked Pose Driver (with Fallbacks)")]
+    public class TrackedPoseDriverWithFallback : TrackedPoseDriver
+    {
+        /// <summary>
+        /// These are the same flags as TrackingState in <seealso cref="TrackedPoseDriver"/> they are repeated here because enum
+        /// TrackingStates is not public in TrackedPoseDriver class (as of Unity.InputSystem 1.8.1.0).
+        /// </summary>
+        [Flags]
+        public enum TDPwithFallbackTrackingStates
```
AMollis, line 32: make this private.

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverWithFallback.cs
```diff
@@ -0,0 +1,168 @@
...
+                rotation = fallbackRotationAction.action.ReadValue<Quaternion>();
+            }
+
+            if (neededToGetFallbackData) //because either position, rotation, or both data were obtained from fallback actions
+            {
+                SetLocalTransformFromFallback(position, rotation, (TDPwithFallbackTrackingStates)fallbackInputTrackingState);
+            }
+        }
+
+        protected virtual void SetLocalTransformFromFallback(Vector3 newPosition, Quaternion newRotation, TDPwithFallbackTrackingStates currentFallbackTrackingState)
```
AMollis, line 140: Make SetLocalTransformFromFallback private for now

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverWithFallback.cs
AMollis: shame we can't reuse the existing `TrackingState`

AMollis (changes requested):
Some more feedback.

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverWithFallback.cs
ms-RistoRK: Done in https://github.com/ms-RistoRK/MixedRealityToolkit-Unity/commit/ad3addc04b06f24c1939dc02320f6517928dcfe5 , resolving.

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverWithFallback.cs
ms-RistoRK: > I don't think you need lines 106 -109.
> 
> Why are you overriding `inputTrackingState` with `FallbackTrackingStateAction`?
> 

I added a comment explaining why.
image

> If inputTrackingState starts out with no valid position or rotation, then we want to use the fallback. However, if you're setting `inputTrackingState` to `FallbackTrackingStateAction`, then lines 123 - 124 and 130-131 will never execute.

ms-RistoRK: done in https://github.com/ms-RistoRK/MixedRealityToolkit-Unity/commit/c902133124685769d3780be832859ae1414c21d1 , resolving.

AMollis (approved):

shaynie (approved):

PR #758, opened 2024-05-30
XRI3 migration step09

ms-RistoRK:
* Added XRI3+ versions of BasicInputTests InputBindingSmoketest, HandMovingSmoketest, InteractableDisabledDuringInteraction, SpawnInteratableOnHand, and UntrackedHandNearInteractions.

### DIFF
--- a/org.mixedrealitytoolkit.input/Tests/Runtime/BasicInputTests.cs
+++ b/org.mixedrealitytoolkit.input/Tests/Runtime/BasicInputTests.cs
@@ -18,12 +18,9 @@ using UnityEngine.XR.Interaction.Toolkit.Interactors;
 using MixedReality.Toolkit.Subsystems;
 
 using HandshapeId = MixedReality.Toolkit.Input.HandshapeTypes.HandshapeId;
-using UnityEngine.InputSystem.XR;
 using System;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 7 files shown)


org.mixedrealitytoolkit.input/Tests/Runtime/BasicInputTests.cs
```diff
@@ -638,7 +639,7 @@ public IEnumerator HandModelHasRequiredFieldsAndAccessors()
         }
 
         // Returns true iff any of the ProximityDetectors in the scene are currently triggered.
```
shaynie, line 641: While you're here, please fix the old typo in the comment.  "iff"

shaynie (approved):

org.mixedrealitytoolkit.input/Tests/Runtime/BasicInputTests.cs
ms-RistoRK: Thanks for catching that.  Not a typo though, 'iff' is the abbreviated version of 'if and only if', although expanding it improves readability.  I expanded all 'iff' abbreviations in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/758/commits/c4270ed3e8eefaeee97dbb9e234da7a24c3dec24 , resolving.

shaynie (approved):

PR #762, opened 2024-06-04
Update GOVERNANCE.md

graceCode100:
Linking SC and Maintainers


GOVERNANCE.md
```diff
@@ -28,15 +28,20 @@ This project may include the following roles. Additional roles may be adopted an
 
 **3.5. Written procedures**. This governance document and other materials documenting this project's development process shall be available to any interested person.
 
-## 4. No Confidentiality
+## 4. Steering Committee
```
AMollis, line 31: Curious why this is needed?  Section `*2.2. Appeal Process*` describes how the SC controls projects.  I suggest we nnot complicate with this messaging

AMollis (changes requested):

GOVERNANCE.md
MaxPalmer-UH: @graceCode100 What's the main aim of this change - to make it clearer to the readers of these documents that the steering committee exists and what its role is?

graceCode100: Hi Max, yes! It's to provide a link between both the maintainers and steering committee repos. Currently, it's not obvious what the steering committee is even though it's referenced briefly in the governance doc and it's not obvious how to get from one to the other from each repo. So i made this update to both the maintainers and steering committee docs to reference each other

keveleigh (approved):

PR #767, opened 2024-06-06
XRI3 migration step14

ms-RistoRK:
* Adding ControllerlessRigSmokeTest
* Making some interactor properties public so that they can be tested.
* Added InputTestUtilities.RigReference to facilitate testing.

### DIFF
--- a/org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
+++ b/org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
@@ -31,7 +31,7 @@ namespace MixedReality.Toolkit.Input
         /// <summary>
         /// Holds a reference to the <see cref="TrackedPoseDriver"/> associated to this interactor if it exists.
         /// </summary>
-        private TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
+        public TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
 
         [Header("Gaze Pinch interactor settings")]
 
@@ -94,7 +94,7 @@ namespace MixedReality.Toolkit.Input
         /// <summary>
         /// The interactor we're using to query potential gaze pinch targets.
         /// </summary>
-        protected XRBaseInputInteractor DependentInteractor { get => dependentInteractor; set => dependentInteractor = value; }
+        public XRBaseInputInteractor DependentInteractor { get => dependentInteractor; set => dependentInteractor = value; }
 
         [SerializeField]
         [Range(0, 1)]
--- a/org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
+++ b/org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
@@ -31,7 +31,7 @@ namespace MixedReality.Toolkit.Input
         /// <summary>
         /// Holds a reference to the <see cref="TrackedPoseDriver"/> associated to this interactor if it exists.
         /// </summary>
-        public TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
+        protected internal TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
 
         [Header("Gaze Pinch interactor settings")]
 
@@ -94,7 +94,7 @@ namespace MixedReality.Toolkit.Input
         /// <summary>
         /// The interactor we're using to query potential gaze pinch targets.
         /// </summary>
-        public XRBaseInputInteractor DependentInteractor { get => dependentInteractor; set => dependentInteractor = value; }
+        protected internal XRBaseInputInteractor DependentInteractor { get => dependentInteractor; set => dependentInteractor = value; }
 
         [SerializeField]
         [Range(0, 1)]
--- a/org.mixedrealitytoolkit.input/Interactors/Poke/PokeInteractor.cs
+++ b/org.mixedrealitytoolkit.input/Interactors/Poke/PokeInteractor.cs
@@ -30,7 +30,7 @@ namespace MixedReality.Toolkit.Input
         /// <summary>
         /// Holds a reference to the <see cref="TrackedPoseDriver"/> associated to this interactor if it exists.
         /// </summary>
-        private TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
+        public TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
 
         [SerializeReference]
         [InterfaceSelector(true)]
--- a/org.mixedrealitytoolkit.input/Interactors/Poke/PokeInteractor.cs
+++ b/org.mixedrealitytoolkit.input/Interactors/Poke/PokeInteractor.cs
@@ -30,7 +30,7 @@ namespace MixedReality.Toolkit.Input
         /// <summary>
         /// Holds a reference to the <see cref="TrackedPoseDriver"/> associated to this interactor if it exists.
         /// </summary>
-        public TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
+        protected internal TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
 
         [SerializeReference]
         [InterfaceSelector(true)]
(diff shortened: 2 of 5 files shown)


org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -94,7 +94,7 @@ public class GazePinchInteractor :
         /// <summary>
         /// The interactor we're using to query potential gaze pinch targets.
         /// </summary>
-        protected XRBaseInputInteractor DependentInteractor { get => dependentInteractor; set => dependentInteractor = value; }
+        public XRBaseInputInteractor DependentInteractor { get => dependentInteractor; set => dependentInteractor = value; }
```
marlenaklein-msft, line 97: Can we make these protected internal instead of public? Not sure we want to expose this

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
```diff
@@ -31,7 +31,7 @@ public class GazePinchInteractor :
         /// <summary>
         /// Holds a reference to the <see cref="TrackedPoseDriver"/> associated to this interactor if it exists.
         /// </summary>
-        private TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
+        public TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
```
AMollis, line 34: I would like to avoid making this public just for testing purposes.  PL=lease let's find a different solution.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
AMollis, line 97: I would like to avoid making this public just for testing purposes.  PL=lease let's find a different solution.

org.mixedrealitytoolkit.input/Interactors/Poke/PokeInteractor.cs
```diff
@@ -30,7 +30,7 @@ public class PokeInteractor :
         /// <summary>
         /// Holds a reference to the <see cref="TrackedPoseDriver"/> associated to this interactor if it exists.
         /// </summary>
-        private TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
+        public TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
```
AMollis, line 33: I would like to avoid making this public just for testing purposes.  PL=lease let's find a different solution.

org.mixedrealitytoolkit.input/Interactors/Ray/MRTKRayInteractor.cs
```diff
@@ -40,7 +40,7 @@ public class MRTKRayInteractor :
         /// <summary>
         /// Holds a reference to the <see cref="TrackedPoseDriver"/> associated to this interactor if it exists.
         /// </summary>
-        private TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
+        public TrackedPoseDriver TrackedPoseDriver => trackedPoseDriver;
```
AMollis, line 43: I would like to avoid making this public just for testing purposes.  PL=lease let's find a different solution.

AMollis (changes requested):

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: You are right, changed to protected internal in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/767/commits/b5240732e09ce6a642878dfb8ef304c92d61a498 , resolving.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: You are right, changed to protected internal in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/767/commits/b5240732e09ce6a642878dfb8ef304c92d61a498 , resolving.

org.mixedrealitytoolkit.input/Interactors/Poke/PokeInteractor.cs
ms-RistoRK: You are right, changed to protected internal in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/767/commits/b5240732e09ce6a642878dfb8ef304c92d61a498 , resolving.

org.mixedrealitytoolkit.input/Interactors/Ray/MRTKRayInteractor.cs
ms-RistoRK: You are right, changed to protected internal in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/767/commits/b5240732e09ce6a642878dfb8ef304c92d61a498 , resolving.

org.mixedrealitytoolkit.input/Interactors/GazePinch/GazePinchInteractor.cs
ms-RistoRK: You are right, changed to protected internal in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/767/commits/b5240732e09ce6a642878dfb8ef304c92d61a498 , resolving.

AMollis (approved):

org.mixedrealitytoolkit.input/Tests/Runtime/BasicInputTestsForControllerlessRig.cs
```diff
@@ -302,6 +347,277 @@ public IEnumerator TrackedHandNearInteractions()
...
+        /// <summary>
+        /// Checks that the Controllerless rig has the right components and that both pre-XRI3+ and XRI3+ input actions are properly referenced.
+        /// The goal of this test is to prevent breaks when the Controllerless rig prefabs or its associated scripts are modified and prefabs are
+        /// not updated properly.  This test also ensures that the Controllerless rig is shipped in the intended configuration.
+        /// </summary>
+        /// <remarks>
+        /// This is an XRI3+ exclusive test.
+        /// </remarks>
+        [UnityTest]
+        public IEnumerator ControllerlessRigSmokeTest()
```
whebertML, line 360: Testing for highly specific configuration/naming/layout for a new rig that is perhaps still in flux and likely to change in the future, especially as we incorporate more XRI directly, seems fragile and likely to need constant modification as we progress.  Is that the idea - to better track changes to the new rig?

ms-RistoRK: Thank you for sharing your thoughts.  One of the reasons is to indeed better keep track changes of the rig but there are also other reasons why I insist on having this test:
* Prevent false positives of the rig working correctly while wrongly configured:  the new XRI3 package has a fallback functionality that if an input action is not found in the TrackedPoseDriver then it falls back to the input action set in the XRController (if present).  Such fallback, and since we are not removing the XRController yet, can give the wrong impression that the Controllerless rig is working properly in an XRI3+ fashion while in reality it is working as the pre-XRI3 rig.  This Unity-test prevents such false positives.
* Prevent regressing to pre-XRI3 configuration.  Even though we keep the XRController for backward compatiblity and to support our customers that are still or want to keep using, we also want to keep aiming and be ready for the moment when Unity decides to remove them in its entirety after the deprecation period.  This Unity-test keeps us in track to what it is going to happen and avoid repeating the effort that has already been invested on.
* Ensure that the input action references in the prefabs are not lost when the scripts are edited, it happened to me a few times as I was making progress in the XRI3 migration.  This Unity-test prevents accidental input action references loss in PRs.
* Re: Using names to find child GameObjects and Input Actions:  I am not too worried about that since names are unlikely to change to prevent breaking changes on our clients that use names in their own apps or forks.  This Unity-test helps prevent re-naming that may break our customers solutions.
* Re: needing constant modification:  Even though I agree this Unity-test does encompass additional modifications, I actually see it as a positive outcome because:
  * Any change to the Controllerless rig should be a well thought change that keeps us aligned to our multiple goals.  This Unity-test helps us think of the impact of such changes.
  * It is easier and faster to immediately address a breaking change or an undetected functionality loss in the PR/commit where it is introduced rather than having to dig in the code for the root-cause when the functionality loss is reported.  This Unity-test helps to immediately take action on a breaking change or a functionality loss rather than later having to dig out a root-cause and implement a fix.

PR #785, opened 2024-06-19
Fix TapToPlace not starting when called before its Start (#784)

### LINKED ISSUE #784: [BUG] `TapToPlace` doesn't start placement when `StartPlacement()` is called just after prefab creation
## Describe the bug

When you call `StartPlacement()` on a `TapToPlace` just after a prefab creation (or `AddComponent<TapToPlace>()`), the TapToPlace isn't being placed.

## To reproduce

1. Add this test to `SolverTapToPlaceTests.cs`
```C#
/// <summary>
/// Verify TapToPlace can start placement when method is called before its own Start.
/// </summary>
[UnityTest]
public IEnumerator TapToPlaceIsBeingPlacedBeforeStart()
{
    // Disable gaze interactions for this unit test;
    InputTestUtilities.DisableGazeInteractor();

    // Set up GameObject with a SolverHandler
    var testObject = GameObject.CreatePrimitive(PrimitiveType.Cube);
    var solverHandler = testObject.AddComponent<SolverHandler>();
    var solver = testObject.AddComponent<TapToPlace>();

    // Disable smoothing so moving happens instantly. This makes testing positions easier.
    solver.Smoothing = false;

    // Set it to track interactors
    solverHandler.TrackedHandedness = Handedness.Both;
    solverHandler.TrackedTargetType = TrackedObjectType.Interactor;
    var lookup = FindObjectUtility.FindAnyObjectByType<ControllerLookup>();
    var leftInteractor = lookup.LeftHandController.GetComponentInChildren<MRTKRayInteractor>();
    var rightInteractor = lookup.RightHandController.GetComponentInChildren<MRTKRayInteractor>();
    solverHandler.LeftInteractor = leftInteractor;
    solverHandler.RightInteractor = rightInteractor;

    int onPlacingStartedCount = 0;
    int onPlacingStoppedCount = 0;
    solver.OnPlacingStarted.AddListener(() => onPlacingStartedCount++);
    solver.OnPlacingStopped.AddListener(() => onPlacingStoppedCount++);

    // Call immediately whithout the TapToPlace Start being called
    solver.StartPlacement();

    yield return RuntimeTestUtilities.WaitForUpdates(1);

    Assert.IsTrue(solver.IsBeingPlaced, "TapToPlace should have started.");
    Assert.AreEqual(1, onPlacingStartedCount, "TapToPlace should have invoked event OnPlacingStarted exactly 1 time.");
    Assert.AreEqual(0, onPlacingStoppedCount, "TapToPlace shouldn't have invoked event OnPlacingStarted.");
}
```
2. Run the test

## Expected behavior

The test succeed, the tap to place is being placed as soon as its Start() method have been called.

- Unity Version 2022.3.7f1
- MRTK Version or Commit [ded4a78](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/ded4a780a4f54c4deeebf5d35e16b85b8745d2d1)

- HoloLens 2

## Additional context

There is a proprety called `placementRequested` already made to fix this usecase, but the DoubleClickTimeout check conflict and cancel the request.
This usecase isn't tested.

anonymous2585:
Fixes #784 

Added non-regression test.

Also corrected comments in tests.

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
@@ -331,9 +331,6 @@ namespace MixedReality.Toolkit.SpatialManipulation
                 return;
             }
 
-            // Get the time of this click action
-            lastTimeClicked = Time.time;
-
             using (StartPlacementPerfMarker.Auto())
             {
                 // Check to see if Start() has been called, if not set placementRequested to true. This will make sure StartPlacement() will be
@@ -347,6 +344,9 @@ namespace MixedReality.Toolkit.SpatialManipulation
                 // Added for code configurability to avoid multiple calls to StartPlacement in a row
                 if (!IsBeingPlaced)
                 {
+                    // Get the time of this click action
+                    lastTimeClicked = Time.time;
+
                     // Store the initial game object layer
                     GameObjectLayer = gameObject.layer;
 
@@ -396,14 +396,15 @@ namespace MixedReality.Toolkit.SpatialManipulation
             {
                 return;
             }
-            // Get the time of this click action
-            lastTimeClicked = Time.time;
 
             using (StopPlacementPerfMarker.Auto())
             {
                 // Added for code configurability to avoid multiple calls to StopPlacementViaPerformedAction in a row
                 if (IsBeingPlaced)
                 {
+                    // Get the time of this click action
+                    lastTimeClicked = Time.time;
+
                     // Change the game object layer back to the game object's layer on start
                     gameObject.layer = GameObjectLayer;
 
--- a/org.mixedrealitytoolkit.core/Tests/TestUtilities/BaseRuntimeTests.cs
+++ b/org.mixedrealitytoolkit.core/Tests/TestUtilities/BaseRuntimeTests.cs
@@ -20,7 +20,7 @@ namespace MixedReality.Toolkit.Core.Tests
         /// Get the target framerate at which tests should run at.
         /// </summary>
         /// <remarks>
-        /// This is used so frame timings to be consistent across various machine types. Thus ensure consistent test behavios.
+        /// This is used so frame timings to be consistent across various machine types. Thus ensure consistent test behaviors.
         /// </remarks>
         protected virtual int TargetFrameRate { get; } = 60;
 
--- a/org.mixedrealitytoolkit.spatialmanipulation/CHANGELOG.md
+++ b/org.mixedrealitytoolkit.spatialmanipulation/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## Unreleased
+
+### Fixed
+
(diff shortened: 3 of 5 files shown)


org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
AMollis, line 1: Update changelog.md with details of fix

org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
```diff
@@ -328,7 +328,7 @@ public void StartPlacement()
             // selected and then immediately unselected. If two calls occur within the
             // double click timeout, then return to prevent an immediate object state switch.
             // Also, check that time is no 0 to allow for auto start functionality.
-            if (Time.time != 0 && (Time.time - lastTimeClicked) < DoubleClickTimeout)
+            if (Time.time != 0 && (Time.time - lastTimeClicked) < DoubleClickTimeout && !placementRequested)
```
AMollis, line 331: Perhaps initializing `lastTimeClicked = float.MaxValue` might be a better fix.

AMollis: Cause theoretically if Time.time is < DoubleClickTimeout, the same bug would occur

org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
AMollis: And check it patch version needs to be updated

AMollis (changes requested):

org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
AMollis, line 331: why not just move `lastTimeClicked = Time.time;` after the `startCalled` change (line 348).   That remove the need to reset placementRequested.   And maybe initialize lastTimeClicked to `float.MaxValue`

Also..IMO, resetting placementRequested makes less sense, given the variable name.  If a reset is preferred...a better name might be `placementRequestedAndPending`

AMollis (changes requested):

org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
anonymous2585: done

org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
anonymous2585: I agree to move the `lastTimeClicked = Time.time;`.

We can't initialize lastTimeClicked to `float.MaxValue`, otherwise `(Time.time - lastTimeClicked) < DoubleClickTimeout` will be `true`, thus the function returns and never starts the tap to place.

AMollis (approved):

PR #794, opened 2024-06-24
Fix for InputSimulator not executing before InteractionManager

ms-RistoRK:


keveleigh (approved):

AMollis (approved):

keveleigh:
@ms-RistoRK Just realized this went into the XRI3 branch. Would we want this change in main too?

ms-RistoRK:
> @ms-RistoRK Just realized this went into the XRI3 branch. Would we want this change in main too?

Good point, yes, we definitively want this in main too.  I'll create a separate PR for that and ping you when ready.

PR #799, opened 2024-06-25
XRI3: Adding a `HandPoseDriver` component that can track poses of devices without a Hand Interaction Profile (Step 1)

AMollis:
# Overview

This is a change for the XRI3 feature branch and is needed to replace functionality that was in the `ArticulatedHandController` which is now deprecated.   In XRI2, the `ArticulatedHandController` would track hand devices that did not have a Hand Interaction Profile.  This new component does something similar, but in an XRI3 supported manner.

- Adding a `HandPoseDriver` component that track hand poses for handle devices without a Hand Interaction Profile

Usage has not been added to the scene yet.  This will come in a future pull request. Also, this has been tested locally, but a full test pass will come with a pull request adding this to the scene.


ms-RistoRK (dismissed):

org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
```diff
@@ -0,0 +1,267 @@
...
+
+            return poseRetrieved;
+        }
+
+        /// <summary>
+        /// Sets the transform that is being driven by the <see cref="TrackedPoseDriver"/>. This will only set requested values, but regardless of tracking state.
+        /// </summary>
+        /// <param name="newPosition">The new local position to possibly set.</param>
+        /// <param name="newRotation">The new local rotation to possibly set.</param>
+        protected virtual void ForceSetLocalTransform(Vector3 newPosition, Quaternion newRotation)
```
keveleigh, line 141: I don't see this method used currently. Is that coming in a future PR, or is this stale?

ericmullis-msft (approved):

org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
```diff
@@ -0,0 +1,266 @@
...
+                // Treat a disabled action as the default None value for the ReadValue call
+                m_trackingState = InputTrackingState.None;
+                return;
+            }
+
+            if (trackingStateAction == null || trackingStateAction.bindings.Count == 0)
+            {
+                // Treat an Input Action Reference with no reference the same as
+                // an enabled Input Action with no authored bindings, and allow driving the Transform pose.
+                m_trackingState = InputTrackingState.Position | InputTrackingState.Rotation;
```
keveleigh, line 204: I'm not sure I follow this logic exactly. In this case, should we drive `m_trackingState` based on whether we're able to get a value from the polyfill?

shaynie (dismissed):

org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
AMollis: doh...This should have been called at line 76.... `SetLocalTransform`   is a base class, and won't apply pose updates if device isn't tracked.

org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
AMollis: I was trying to reproduce logic used in `UnityEngine.InputSystem.XR.TrackedPoseDriver`... see this in `com.unity.inputsystem@1.8.1\InputSystem\Plugins\XR\TrackedPoseDriver.cs`

```cs
        void ReadTrackingState()
        {
            var trackingStateAction = m_TrackingStateInput.action;
            if (trackingStateAction != null && !trackingStateAction.enabled)
            {
                // Treat a disabled action as the default None value for the ReadValue call
                m_CurrentTrackingState = TrackingStates.None;
                return;
            }

            if (trackingStateAction == null || trackingStateAction.m_BindingsCount == 0)
            {
                // Treat an Input Action Reference with no reference the same as
                // an enabled Input Action with no authored bindings, and allow driving the Transform pose.
                m_CurrentTrackingState = TrackingStates.Position | TrackingStates.Rotation;
                return;
            }

           ....
```

My thought is that it would be odd if `HandPoseDriver` determined tracking state differently than the base component.  Also, it's super unfortunate that the base does not expose `m_CurrentTrackingState` to subclasses.

keveleigh: Sounds good to me! Thanks for the explanation!

keveleigh (approved):

ericmullis-msft (approved):

PR #800, opened 2024-06-25
XRI3: Adding `PinchInputReader` to allow pinch selection and pinch progress on devices without a Hand Interaction Profile (Step 1)

AMollis:
# Overview 
This is a change for the XRI3 feature branch and is needed to replace functionality that was in the `ArticulatedHandController` which is now deprecated.   In XRI2, the `ArticulatedHandController` would track pinch state for hand devices that did not have a Hand Interaction Profile.  This new component does something similar, but in an XRI3 supported manner.

- Adding `PinchInputReader` to allow pinch selection and pinch progress on devices without a Hand InteractionProfile 

Usage has not been added to the scene yet.  This will come in a future pull request.   Also, this has been tested locally, but a full test pass will come with a pull request adding this to the scene.


shaynie (approved):

ms-RistoRK (dismissed):

ericmullis-msft (dismissed):

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
```diff
@@ -0,0 +1,441 @@
...
+        /// </remarks>
+        /// <seealso cref="SelectActionValue"/>
+        public InputActionProperty SelectAction
+        {
+            get => selectAction;
+            set => SetInputActionProperty(ref selectAction, value);
+        }
+
+        [SerializeField, Tooltip("The Input System action to read values for selecting an Interactable. If not defined or if a controller is not attached to this property, the selection value will be driven by the IHandsAggregatorSubsystem subsystem's pinch amount.")]
+        InputActionProperty selectActionValue = new InputActionProperty(new InputAction("Select Value", expectedControlType: "Axis"));
```
keveleigh, line 68: nit: missing explicit access modifier

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
```diff
@@ -0,0 +1,441 @@
...
+        }
+
+        #endregion Serialized Fields
+
+        #region Private Fields
+
+        FallbackState m_fallbackState;
+        InputTrackingState m_trackingState;
+        bool m_firstUpdate = true;
+        bool m_trackingActionBound = false;
```
keveleigh, line 118: nit: missing explicit access modifiers

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
```diff
@@ -0,0 +1,441 @@
...
+    /// This is a workaround for device's without interaction profiles for hands. Once universal hand interaction profiles are available,
+    /// this class will be removed.
+    /// </remarks>
+    [DefaultExecutionOrder(XRInteractionUpdateOrder.k_XRInputDeviceButtonReader)]
+    public class PinchInputReader : MonoBehaviour, IXRInputButtonReader
+    {
+        /// <summary>
+        /// The state of the pinch input reader when <see cref="InputActionProperty"/> is not set or not bound to a control.
+        /// </summary>
+        private struct FallbackState
```
keveleigh, line 29: Should/could we use Unity's `InteractionState` struct instead of defining our own? It also comes with helpers, like `SetFrameState`, which sets all four values in one method call.

AMollis: I don't think so.  XRI3 has lightly deprecated `InteractionState` ... and now states the following:

```cs
    /// <remarks>
    /// Use of this class should be avoided outside of the context of playback and recording.
    /// Instead, use <see cref="XRBaseInputInteractor.LogicalInputState"/> when possible.
    /// </remarks>
```

AMollis: And `LogicalInputState` looks like too much for this case.

Sort of related to this comment....Unity recommended implementing `IXRInputButtonReader` to handle this "No Hand Interaction Profile".  The pattern here is following the pattern in the XRI3 hand input sample scene.

keveleigh: Same question then, but "Should/could we use Unity's _LogicalInputState_ struct instead of defining our own?" 😉

keveleigh: > And LogicalInputState looks like too much for this case.

Beat me to my follow-up haha!

AMollis: @keveleigh ... So are you good with this? :)

keveleigh: Ah yeah, that's fine!

shaynie (approved):

PR #802, opened 2024-06-25
Fix MRTKRayInteractor always using hand palm direction in isHoverActive

### LINKED ISSUE #801: [BUG] MRTKRayInteractor always attempts verifying palm facing direction even for non-handed controllers.
## Describe the bug

In the feature/XRI3 branch, the updated MRTKRayInteractor always attempts verifying the hand palm facing direction, even if the controller is non-handed.

## To reproduce

For controllers that aren't driven by hand tracking, but hand tracking is also active (held tracked stylus, or Magic Leap Controller, for instance), this will cause the ray interactor to flicker on and off based on the hand's palm facing direction.  

## Expected behavior

For general tracked controllers, their ray interactor should not be influenced by the hand's palm facing direction.

- Unity Version 2022.3.27f1
- MRTK Version or Commit [916fd49](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/916fd495c6776dd049fccfde1f7eff423d05bdbb)

- HoloLens 2
- Magic Leap
- OpenXR

whebertML:
* Fixes #801


org.mixedrealitytoolkit.input/Interactors/Ray/MRTKRayInteractor.cs
```diff
@@ -232,7 +232,8 @@ public override bool isHoverActive
                         }
 #pragma warning restore CS0612
 #pragma warning restore CS0618
-                        else
+                        // Don't attempt palm check if the interactor is not driven by a hand (i.e. non-handed tracked controllers).
```
AMollis, line 235: nit: this new if statement can probably be merged with the next one

```cs
else if (handedness != InteractorHandedness.None &&
XRSubsystemHelpers.HandsAggregator?.TryGetPalmFacingAway(handedness.ToXRNode(), out isPalmFacingAway) ?? true)
{
   hoverActive &= isPalmFacingAway;
}
```

AMollis: Actually...this should probably be this...that way is there is no hand aggreator we fallback to base behavior

```cs
else if (handedness != InteractorHandedness.None &&
XRSubsystemHelpers.HandsAggregator != null && 
XRSubsystemHelpers.TryGetPalmFacingAway(handedness.ToXRNode(), out isPalmFacingAway))
{
   hoverActive &= isPalmFacingAway;
}
```

AMollis: I wish I caught this in a previous review

AMollis (changes requested):
Like your change...but can you clean up the old change :)

org.mixedrealitytoolkit.input/Interactors/Ray/MRTKRayInteractor.cs
whebertML: The initial if, involving forcedDeprecatedInput and ArticulatedHandController, will also fall into the if block due to the null coalescing operator ?? true if there is no HandsAggregator.  But this is how MRTKRayInteractor behaved prior to the XRI3 changes, so the current is continuing that behavior and not changing it.  Would you want both changed, or leave using null coalescing op?

```
if (forceDeprecatedInput &&
    xrController is ArticulatedHandController handController &&
    (XRSubsystemHelpers.HandsAggregator?.TryGetPalmFacingAway(handController.HandNode, out isPalmFacingAway) ?? true))
{
        hoverActive &= isPalmFacingAway;
}
```

AMollis: hmmm good point.  

Ok disregard my second comment....can you at least fix me nit :)

```cs
else if (handedness != InteractorHandedness.None &&
XRSubsystemHelpers.HandsAggregator?.TryGetPalmFacingAway(handedness.ToXRNode(), out isPalmFacingAway) ?? true)
{
   hoverActive &= isPalmFacingAway;
}
```

whebertML: @AMollis Sounds good, will do, thanks for review.

whebertML: Else if merged

AMollis (approved):

org.mixedrealitytoolkit.input/Interactors/Ray/MRTKRayInteractor.cs
```diff
@@ -232,12 +232,11 @@ public override bool isHoverActive
                         }
 #pragma warning restore CS0612
 #pragma warning restore CS0618
-                        else
+                        // Attempt palm facing away check if the interactor is associated with a hand.
+                        else if (handedness != InteractorHandedness.None &&
+                            (XRSubsystemHelpers.HandsAggregator?.TryGetPalmFacingAway(handedness.ToXRNode(), out isPalmFacingAway) ?? true))
```
keveleigh, line 237: should this be `?? false`? If `HandsAggregator` is null, then we won't have a meaningful value for `isPalmFacingAway`

whebertML: @keveleigh @AMollis I would tend to agree, but I believe @ms-RistoRK was leaving the [existing logic](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/main/org.mixedrealitytoolkit.input/Interactors/Ray/MRTKRayInteractor.cs#L161) of `?? true` in place - not sure why that was originally chosen if HandsAggregator was null, though?

PR #808, opened 2024-06-26
XRI3 migration fixing scenes reapplying new rig

ms-RistoRK:
* Restored all .unity scene files with a copy from main branch
* In each scene replace obsolete MRTK Rig with new controllerless rig.

Note:  To produce this PR the steps followed were:
1 - Overwrite all .unity scene files in feature/XRI3 with a copy of the same files from main branch.
2 - Opened each scene and removed the obsolete MRTK rig.
2.1 - Added the new controllerless MRTK rig to the scene.
3 - Restored lost scenes (DisableInteractorsExample, SpatialMouseSample, SpatialMappingExample, TapToPlaceExample, and EyeTrackingPositioningExample) references because of obsolete MRTK rig replacement.
4 - Re-applied previous changes to ObsoleteHandInteractionExamples scene description panel.

Additional manual testing done:
* Sideloaded a build to HL2 and tested all scenes, results: 
  * Core functionality tested in all scenes, no core functionality impairment detected.
  * No broken assets detected.

Note: This PR is difficult to appreciate the scene changes so to review it I recommend to clone to a local repo and open Unity Editor.  Open each scene and expand the MRTK Rig GameObject and verify that the rig's MRTKRightHandController child has an emptied XRController and a Tracked Pose Driver component.  This would indicate that the scene does have the new controllerless MRTK Rig.  For example:
image

### DIFF
--- a/UnityProjects/MRTKDevTemplate/Assets/Scenes/Audio/AudioLoFiExample.unity
+++ b/UnityProjects/MRTKDevTemplate/Assets/Scenes/Audio/AudioLoFiExample.unity
@@ -104,7 +104,7 @@ NavMeshSettings:
   serializedVersion: 2
   m_ObjectHideFlags: 0
   m_BuildSettings:
-    serializedVersion: 3
+    serializedVersion: 2
     agentTypeID: 0
     agentRadius: 0.5
     agentHeight: 2
@@ -117,7 +117,7 @@ NavMeshSettings:
     cellSize: 0.16666667
     manualTileSize: 0
     tileSize: 256
-    buildHeightMesh: 0
+    accuratePlacement: 0
     maxJobWorkers: 0
     preserveTilesOutsideBounds: 0
     debug:
@@ -154,6 +154,7 @@ RectTransform:
   m_ConstrainProportionsScale: 0
   m_Children: []
   m_Father: {fileID: 1231290232}
+  m_RootOrder: 2
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_AnchorMin: {x: 0, y: 1}
   m_AnchorMax: {x: 0, y: 1}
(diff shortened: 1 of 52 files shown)


UnityProjects/MRTKDevTemplate/ProjectSettings/boot.config
AMollis, line 1: don't add this file...

AMollis (approved):

UnityProjects/MRTKDevTemplate/ProjectSettings/boot.config
AMollis: unless we know for sure its purpose

ms-RistoRK: I think we should add it because it is located under ProjectSettings folder and according to [Unity's manual](https://docs.unity3d.com/2021.2/Documentation/Manual/UnityCollaborateIgnoreFiles.html) they mention that we should not exclude the **Project Settings** folder as in:
image
As for its purpose the only reference I could find in the official documentation is that it is generated by [BuildPipeline.WriteBootConfig](https://docs.unity3d.com/ScriptReference/BuildPipeline.WriteBootConfig.html) which doesn't give a lot of details about the file.  The lack of documentation seems to be intentional though, a Unity developer [posted in this forum](https://forum.unity.com/threads/info-or-docs-on-boot-config.971457/) that it is undocumented intentionally.
image
I would say we should add it, What do you think?

ms-RistoRK: Update:  I also found some other people mentioned it is regenerated every time and it changes between Editor versions and platform, So it is not needed to be source controlled, I'll remove it.

ms-RistoRK: > If we are going to omit from source control, should we add it to the project's .gitignore?

That's a good question.  I would say no (imo) because we already have it as part of the repo and it is already source-controlled:
image
If someone forks MRTK3 and they actually need the file source-controlled for their solutions we would be causing them a headache having them trying to figure out why their specific Editor+Platform configuration is no longer properly source-controlled.  What do you think?

AMollis (approved):

PR #810, opened 2024-06-27
[XRI3] Update MRTK3 Input Hand Simulation to better work with XRI3

AMollis:
# Overview

- Refactoring hand controller updates in test framework, so to work with XRI3 update loop.
- Also refactoring how the rig is chosen in the test cases. By default, we can use the new rig.

# Details

This change allows simulate hand updates, during unit tests, to be executed at a particular time within the update loop. For example, the caller may want to hand's updates to tick immediately after the input system has performed its updates.

Controlling when update is call is important since the input system is particular on when and input action is "performed this frame".  Updates to the hand controller can result in a selection action to be trigger. In order for an action to work, the update times of an input action's `InputActionState.TriggerState.lastPerformedInUpdate` need to match exactly with the frame counter the action was actually updated in, see `InputSystem.LowLevel.InputUpdate.s_UpdateStepCount`. This counter, `InputUpdate.s_UpdateStepCount`, is updated during `InputSystem.onBeforeUpdate`and before `XRInteractionManager` is updated. This means that the hand controller updates need to happen during `InputSystem.onAfterUpdate` so that `XRInteractionManager` is able see a proper update of the hand controller's selection.


ms-RistoRK (approved):
Approved with minor suggestions.

shaynie (changes requested):

AMollis:
@shaynie and @ms-RistoRK  resolved your feedback

shaynie (approved):

shaynie (approved):

ms-RistoRK (approved):

PR #811, opened 2024-06-27
[XRI3] Adding tracked pose driver extensions

AMollis:
# Overview

Adding tracked pose driver extensions. These will help improve "tracking status" queries across interactors.  It also exposes a way for callers to obtain the "cached tracking status" on the HandPoseDriver.

### DIFF
--- a/org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
+++ b/org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
@@ -46,7 +46,7 @@ namespace MixedReality.Toolkit.Input
         /// </remarks>
         public static InputTrackingState GetInputTrackingState(this TrackedPoseDriver driver)
         {
-            // If the driver is a HandPoseDriver, return the cached value, istead of hitting the overhead of querying the action.
+            // If the driver is a HandPoseDriver, return the cached value, instead of hitting the overhead of querying the action.
             if (driver is HandPoseDriver handPoseDriver)
             {
                 return handPoseDriver.CachedTrackingState;
--- a/org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
+++ b/org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
@@ -61,7 +61,7 @@ namespace MixedReality.Toolkit.Input
         /// </summary>
         /// <remarks>
         /// If the <see cref="TrackedPoseDriver"/> has no tracking state action or the action has no bindings, it will return `<see cref="InputTrackingState.Position"/> |
-        /// <see cref="InputTrackingState.Rotation"/>`. If the action is enabled, it will return `<see cref="InputTrackingState.None"/>`. If the action has controls, it will return the value of the action.
+        /// <see cref="InputTrackingState.Rotation"/>`. If the action is disabled, it will return `<see cref="InputTrackingState.None"/>`. If the action has controls, it will return the value of the action.
         /// </remarks>
         internal static InputTrackingState GetInputTrackingStateNoCache(this TrackedPoseDriver driver)
         {
--- a/org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
+++ b/org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
@@ -68,7 +68,7 @@ namespace MixedReality.Toolkit.Input
             // Note, that the logic in this class is meant to reproduce the same logic as the base. The base
             // `TrackedPoseDriver` also sets the tracking state in a similar manner. Please see 
             // `TrackedPoseDriver::ReadTrackingState`. Replicating this logic in a subclass is not ideal, but it is
-            // necessary since the base class does not expose the tracking static update logic.
+            // necessary since the base class does not expose its tracking status field.
 
             var trackingStateAction = driver.trackingStateInput.action;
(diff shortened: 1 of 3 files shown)


org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+        public static bool TryGetTrackingState(this TrackedPoseDriver driver, out InputTrackingState state)
+        {
+            state = InputTrackingState.None;
+            var trackingStateAction = driver.trackingStateInput.action;
+            if (trackingStateAction == null || trackingStateAction.bindings.Count == 0)
+            {
+                return false;
+            }
+
+            if (trackingStateAction.enabled)
```
whebertML, line 26: Should this be `if (!trackingStateAction.enabled)`?

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+            return GetInputTrackingStateNoCache(driver);
+        }
+
+
+        /// <summary>
+        /// Gets the tracking state of the <see cref="TrackedPoseDriver"/>, avoid reading value for internal caches.
+        /// </summary>
+        /// <remarks>
+        /// If the <see cref="TrackedPoseDriver"/> has no tracking state action or the action has no bindings, it will return `<see cref="InputTrackingState.Position"/> |
+        /// <see cref="InputTrackingState.Rotation"/>`. If the action is enabled, it will return `<see cref="InputTrackingState.None"/>`. If the action has controls, it will return the value of the action.
```
whebertML, line 64: "If the action is `disabled`, it will return InputTrackingState.None?

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// Gets the tracking state of the <see cref="TrackedPoseDriver"/>.
+        /// </summary>
+        /// <remarks>
+        /// If the <see cref="TrackedPoseDriver"/> has no tracking state action or the action has no bindings, it will return `<see cref="InputTrackingState.Position"/> |
+        /// <see cref="InputTrackingState.Rotation"/>`. If the action is enabled, it will return `<see cref="InputTrackingState.None"/>`. If the action has controls, it will return the value of the action.
+        /// </remarks>
+        public static InputTrackingState GetInputTrackingState(this TrackedPoseDriver driver)
+        {
+            // If the driver is a HandPoseDriver, return the cached value, istead of hitting the overhead of querying the action.
```
ms-RistoRK, line 49:
```suggestion
            // If the driver is a HandPoseDriver, return the cached value, instead of hitting the overhead of querying the action.
```

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
ms-RistoRK, line 64:
```suggestion
        /// <see cref="InputTrackingState.Rotation"/>`. If the action is disabled, it will return `<see cref="InputTrackingState.None"/>`. If the action has controls, it will return the value of the action.
```

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
ms-RistoRK: I think this is correct as it is because TrackedPoseDriver::ReadTrackingState() `"// Treat a disable action as the default None value for the ReadValue call"` as in:
image
Meaning that a disabled `trackingStateAction` can be queried for tracking state and the result be InputTrackingState.None and since this is a `TryGetTrackingState` method then this logic will return true (can be queried) and the value will be InputTrackingState.None.  @AMollis Do you concur?

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+            state = (InputTrackingState)trackingStateAction.ReadValue<int>();
+            return true;
+        }
+
+        /// <summary>
+        /// Gets the tracking state of the <see cref="TrackedPoseDriver"/>.
+        /// </summary>
+        /// <remarks>
+        /// If the <see cref="TrackedPoseDriver"/> has no tracking state action or the action has no bindings, it will return `<see cref="InputTrackingState.Position"/> |
+        /// <see cref="InputTrackingState.Rotation"/>`. If the action is enabled, it will return `<see cref="InputTrackingState.None"/>`. If the action has controls, it will return the value of the action.
```
whebertML, line 45:
```suggestion
/// <see cref="InputTrackingState.Rotation"/>`. If the action is disabled, it will return `<see cref="InputTrackingState.None"/>`. If the action has controls, it will return the value of the action.
```

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
AMollis: Ooops!!!  typo will fix

ms-RistoRK (approved):

AMollis:
@whebertML .... I've addressed your change requests

shaynie (approved):

PR #812, opened 2024-06-27
[XRI3] Fix HandPoseDrive Dispose Handling

AMollis:
# Overview

Fixing HandPoseDrive dispose handling.  The old code failed when running in unit tests, because old object was trying to access the disposed backing-native object.  

NOTE, having to already disposable of this subclass this way is not ideal.  This is code is here as a workaround since the base does not make "OnDisable" virtual.

### DIFF
--- a/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
+++ b/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
@@ -186,7 +186,7 @@ namespace MixedReality.Toolkit.Input
         /// </summary>
         private bool HandleDisablement()
         {
-            // If backing native object has been destoryed (this == null) or component is
+            // If backing native object has been destroyed (this == null) or component is
             // disabled, we should unbind the tracking state updates.
             if (this == null || !isActiveAndEnabled || !Application.isPlaying)
             {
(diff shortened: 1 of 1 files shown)


shaynie (dismissed):

org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
```diff
@@ -184,11 +184,18 @@ private void OnFirstUpdate()
         /// The base class has not made OnDisable virtual, so we need to check for disablement in
         /// tracking state callbacks. If base ever make OnDisable virtual, we can unbind in OnDisable instead.
         /// </summary>
-        private void HandleDisablement()
+        private bool HandleDisablement()
         {
-            if (!isActiveAndEnabled || !Application.isPlaying)
+            // If backing native object has been destoryed (this == null) or component is
```
ms-RistoRK, line 189:
```suggestion
            // If backing native object has been destroyed (this == null) or component is
```

ms-RistoRK (approved):
Just one minor nit suggestion.

shaynie (approved):

ms-RistoRK (approved):

PR #822, opened 2024-06-29
[XRI3] Updating `ObjectManipulator` so to not rely on obsolete XRI controllers. 

AMollis:
# Overview

Updating `ObjectManipulator` so to not rely on obsolete XRI controllers. 

This required introducing a new interactor interface, `ITrackedInteractor`.  This interface provides a way for the `ObjectManipulator` to obtain the transform of the tracked input device.

### DIFF
--- a/org.mixedrealitytoolkit.core/Interactors/ITrackedInteractor.cs
+++ b/org.mixedrealitytoolkit.core/Interactors/ITrackedInteractor.cs
@@ -11,7 +11,7 @@ namespace MixedReality.Toolkit
     public interface ITrackedInteractor
     {
         /// <summary>
-        /// Get the interactor's parent whoee pose is backed by a tracked input device.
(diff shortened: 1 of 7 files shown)


org.mixedrealitytoolkit.core/Interactors/ITrackedInteractor.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace MixedReality.Toolkit
+{
+    /// <summary>
+    /// Represents an interactor whose parent pose is backed by a tracked input device.
+    /// </summary>
+    public interface ITrackedInteractor
+    {
+        /// <summary>
+        /// Get the interactor's parent whoee pose is backed by a tracked input device.
```
AMollis, line 14:
``` suggestion
whose
```

shaynie (approved):

PR #826, opened 2024-07-02
[XRI3] Updating XRI3 rig to use new "HandPoseDrive" and "PinchInputReader" so to support devices without a Hand Interaction profile

AMollis:
# Overview

 Updating XRI3 rig to use new `HandPoseDrive` and `PinchInputReader` so to support devices without a Hand Interaction profile.

I also have to update the `TapToPlace` solver so it worked with custom `XRInputButtonReaders`.  That is. `PinchInputReader` is a custom `XRInputButtonReaders`, and without this update, the `TapToPlace` behaviour would not work with the new rig.

Also, in this change:

- Disabling some see-it-say-label tests on batch mode.  These tests actually always fail to work in batch mode, since batch mode doesn't have speech services available.
- Changing `PinchInputReader` to cache "valid" action states.  To help improve performance
- Changing `PinchInputReader` with a safer way of unregistering from action events
- Adding an assert in the input test helpers, to catch simultaneous rig creations.

The rig changes are as follows:

1. First screenshot shows the `PinchInputReaders` being added to the rig for the left hand
pr-0

2. Second screenshot shows the `PinchInputReaders` being added to the rig for the right hand.
pr-0-right

3. Next is setting up poke interactor with the `PinchInputReaders` (and explicitly setting the tracked pose driver).
pr-1

4. Setting up the far ray interactor with the `PinchInputReaders` (and explicitly setting the tracked pose driver).
pr-2

5. Setting up the grab interactor with the `PinchInputReaders` (and explicitly setting the tracked pose driver).
pr-3

6. Setting up the gaze pinch interactor with the `PinchInputReaders` (and the tracked pose driver was already set).
pr-4

7. Swapping the old `TrackedPoseDriver` with the extended `HandPoseDriver` tracked pose driver class.
pr-5-pose-driver

8. Moving the `InteractionManager` game object to the top of the rig prefab, so to match the old rig layout.
pr-6-move-interaction-manager-to-top

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
@@ -431,7 +431,7 @@ namespace MixedReality.Toolkit.SpatialManipulation
         {
             using (SolverUpdatePerfMarker.Auto())
             {
-                // Stop stop placement if a select action is performed this frame
+                // Stop placement if a select action is performed this frame
                 if (InteractorSelectPerformedThisFrame())
                 {
                     StopPlacement();
--- a/org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
@@ -538,7 +538,7 @@ namespace MixedReality.Toolkit.SpatialManipulation
         }
 
         /// <summary>
-        /// Get if an interactor's selcect button was performed this frame.
+        /// Get if an interactor's select button was performed this frame.
         /// </summary>
         private bool InteractorSelectPerformedThisFrame()
         {
--- a/org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
@@ -6,7 +6,6 @@ using Unity.Profiling;
 using UnityEngine;
 using UnityEngine.Events;
 using UnityEngine.InputSystem;
-using UnityEngine.InputSystem.XR;
 using UnityEngine.XR.Interaction.Toolkit;
 using UnityEngine.XR.Interaction.Toolkit.Inputs.Readers;
 using UnityEngine.XR.Interaction.Toolkit.Interactors;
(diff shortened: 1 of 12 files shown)


org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
```diff
@@ -429,10 +431,15 @@ public override void SolverUpdate()
         {
             using (SolverUpdatePerfMarker.Auto())
             {
+                // Stop stop placement if a select action is performed this frame
```
ms-RistoRK, line 434:
```suggestion
                // Stop placement if a select action is performed this frame
```

org.mixedrealitytoolkit.spatialmanipulation/Solvers/TapToPlace.cs
```diff
@@ -530,6 +537,27 @@ protected virtual void SetRotation()
             }
         }
 
+        /// <summary>
+        /// Get if an interactor's selcect button was performed this frame.
```
ms-RistoRK, line 541:
```suggestion
        /// Get if an interactor's select button was performed this frame.
```

shaynie:
Consistent test failures in the PR build - please review whether the implementation or the tests needs modification.

shaynie (changes requested):
Please address test failures

AMollis:
> Please address test failures

Fixed

org.mixedrealitytoolkit.uxcomponents/Tests/Runtime/SeeItSayItCanvasTests.cs
```diff
@@ -86,6 +86,7 @@ public IEnumerator TestLabelEnabledOnHover()
             yield return hand.Show(Vector3.one);
             yield return RuntimeTestUtilities.WaitForUpdates();
             yield return hand.MoveTo(testButton.transform.position);
+
```
whebertML, line 89: This file only seems to have newline changes, intentional?

ms-RistoRK: Improves readability :)

ms-RistoRK (approved):

shaynie (approved):

PR #834, opened 2024-07-10
Fix RC pipeline serviceconnection

shaynie:


shaynie:
Abandoned in favor of Marlena's

PR #845, opened 2024-07-12
Fixing issues with RiggedHandMeshVisualizer visualizing select value

whebertML:
Fixes the RiggedHandMeshVisualizer visualizing select value in both new XRI3 controllerless rig and the legacy obsolete rig.

* Adds an ISelectInputVisualizer interface, providing a means to set an XRInputButtonReader to represent selection values to be incorporated by a visualizer.
* Makes RiggedHandMeshVisualizer implement ISelectInputVisualizer so select input can be set externally on instantiation.
* Updates HandModel with a select input property that can be set on the instantiated hand model when implementing ISelectInputVisualizer.
* Updates the MRTK LeftHand Controller.prefab to set HandModel with object reference to the PinchInputReader for select input (used by right hand variant as well).
* Fixes an issue with RiggedHandMeshVisualizer utilizing legacy XRBaseController selectInteractionState as a fallback.

Modifications to the new XRI3 controllerless rig, sets the HandModel select input reader to the existing PinchInputReader in both the left and right hand prefabs.

image
image

And when the hand model is instantiated, the select input linkage is made to properly visualize:

image

Also fixes it working for the legacy [Obsolete] rig:

image


org.mixedrealitytoolkit.input/Experimental/XRI3/HandModel.cs
```diff
@@ -72,6 +82,12 @@ protected virtual void Start()
             if (ModelPrefab != null)
             {
                 model = Instantiate(ModelPrefab, ModelParent);
+
+                // Set the select input reader for the model if it implements ISelectInputVisualizer
+                if (selectInput != null && model != null && model.TryGetComponent(out ISelectInputVisualizer selectInputVisualizer))
+                {
+                    selectInputVisualizer.SelectInput = selectInput;
```
AMollis, line 89: Assert if select input is not set

AMollis (approved):
Seems good to me.

Minor feedback

org.mixedrealitytoolkit.input/Experimental/XRI3/HandModel.cs
```diff
@@ -56,6 +57,15 @@ public class HandModel : MonoBehaviour
         /// <remarks>Expected to be XRNode.LeftHand or XRNode.RightHand.</remarks>
         public XRNode HandNode => handNode;
 
+        [SerializeField, Tooltip("The select input reader to be used by the hand model prefab when implementing ISelectInputVisualizer.")]
```
ms-RistoRK, line 60: As per https://learn.microsoft.com/en-us/windows/mixed-reality/mrtk-unity/mrtk2/contributing/coding-guidelines?view=mrtkunity-2022-05#do-10
```suggestion
        [SerializeField, Tooltip("The XRInputButtonReader representing selection values to be used by the hand model prefab when implementing ISelectInputVisualizer.")]
```

ms-RistoRK (dismissed):
LGTM, just one minor suggestion.

org.mixedrealitytoolkit.input/Experimental/XRI3/HandModel.cs
whebertML: Debug.Assert added, resolving.

AMollis (approved):

PR #852, opened 2024-07-16
Fix an issue with PinchInputReader's TryGetValue when using fallback.

whebertML:
* Previously, TryGetValue would only return a value successfully when using fallback and pinch was being performed, not during the whole range of pinch values as expected.


AMollis (approved):
Thanks for fixing this :)

PR #855, opened 2024-07-16
Allow the HandModel ModelPrefab property to be set at runtime.

whebertML:
* This aligns with the functionality of the deprecated ArticulatedHandController's ModelPrefab, which allowed runtime setting, that this HandModel is meant to replace.

### DIFF
--- a/org.mixedrealitytoolkit.input/Controllers/HandModel.cs
+++ b/org.mixedrealitytoolkit.input/Controllers/HandModel.cs
@@ -17,14 +17,18 @@ namespace MixedReality.Toolkit.Input
     {
         #region Properties
 
-        [SerializeField, Tooltip("The prefab of the MRTK Controller to show that will be automatically instantitated by this behaviour.")]
+        [SerializeField, Tooltip("The prefab of the MRTK Controller to show that will be automatically instantiated by this behavior.")]
         private Transform modelPrefab;
 
         /// <summary>
-        /// The prefab of the model to show that will be automatically instantitated by this <see cref="MonoBehaviour"/>.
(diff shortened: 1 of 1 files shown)


AMollis (approved):

org.mixedrealitytoolkit.input/Controllers/HandModel.cs
```diff
@@ -17,14 +17,18 @@ public class HandModel : MonoBehaviour
...
         /// <summary>
-        /// The prefab of the model to show that will be automatically instantitated by this <see cref="MonoBehaviour"/>.
+        /// The prefab of the model to show that will be automatically instantiated by this <see cref="MonoBehaviour"/>.
         /// </summary>
         /// <remarks>Expected to be XRNode.LeftHand or XRNode.RightHand.</remarks>
-        public Transform ModelPrefab => modelPrefab;
+        public Transform ModelPrefab
+        {
+            get => modelPrefab;
+            set => modelPrefab = value;
```
keveleigh, line 30: Should setting this reinitialize the currently loaded model?

whebertML: @keveleigh Good question.  I essentially just copied the behavior of the now deprecated XRBaseController's ModelPrefab, that this HandModel is somewhat replacing (code from XRBaseController below):

```
        public Transform modelPrefab
        {
            get => m_ModelPrefab;
            set => m_ModelPrefab = value;
        }
```

So I expect the intention of XRBaseController is that you must set it during Awake(), before the model is instantiated during Start() - so HandModel now matches that expectation.

keveleigh: Oh good call, yeah, that makes sense to me! Thanks!

PR #858, opened 2024-07-17
Adding hand "squeeze" binding for MRTK3's "selectValue" action

AMollis:
# Overview

In the XRI3 update, HoloLens was incorrectly fallbacking back to using the joint positions to determine the selection value, instead of obtaining the value from the Microsoft's Hand Interaction OpenXR Profile.  This is fixed by adding "<XRController>{LeftHand}/squeeze" and "<XRController>{RightHand}/squeeze" binding for MRTK3's "selectValue" action.

mrtk3-pr1

mrtk3-pr2


MaxPalmer-UH:
I’ll take a look at this tomorrow. Thanks!

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

ms-RistoRK (approved):

org.mixedrealitytoolkit.input/Assets/Input Actions/MRTK Default Input Actions.inputactions
```diff
@@ -529,6 +529,17 @@
                     "isComposite": false,
                     "isPartOfComposite": false
                 },
+                {
+                    "name": "",
+                    "id": "088cece3-47b8-4235-8748-ae126d30e43c",
+                    "path": "<XRController>{LeftHand}/squeeze",
```
keveleigh, line 535: Should these be `select` instead of `squeeze`? I believe that's what maps to an airtap, as opposed to a grab/grasp action

keveleigh: Actually, this XRI3 rework might be a great opportunity to finally split out `select` and `squeeze` into distinct actions mapped to distinct interactors. In the XRController world, we were largely restricted to one action (with multiple mappings) for all interactors of the same hand, but in reality, the action required for near grab should be different than a far airtap. It looks like we have an opportunity to fix that in XRI3 with the button readers.

AMollis: Please open a new item to rework XRI3 Far and Near to use different actions references.   This seems different then the issue that I'm trying to solve here.

I'm trying to match the behavior of MRTK3/XRI2.   In MRTK3/XRI2, it only had a single "select" action, not a "selectValue" action.  The "select" binding in MRTK3/XRI2 was bound to "[squeezePressed](https://docs.unity3d.com/Packages/com.unity.xr.openxr@1.9/api/UnityEngine.XR.OpenXR.Features.Interactions.MicrosoftHandInteraction.HoloLensHand.squeezePressed.html)"  

I could change to "select", but was trying to keep the same behavior as MRTK3/XRI2....thoughts?

AMollis: Come to think of being bound to "squeeze" might be why "Max" is seeing the weird behavior Max is seeing

AMollis: I feel like for this preview release, keeping the same behavior as MRTK3/XRI2 is preferred.  We can revisit with a future change.   Mapping near and far to different actions is not hard in XRI3 but would like to reduce risk in getting the first preview out.

keveleigh: > This seems different than the issue I'm trying to solve here.

Yeah, good point, and agreed that maintaining parity for this initial preview release sounds like a good plan. Did we lose an action reference or something?

I'll also file an issue for investigating replacing our `IPoseSource`s with tracked pose drivers, since that also seems like a potential improvement with XRI3. And adding action references to the cross-vendor hand interaction profile would be a good improvement as well.

keveleigh (approved):

Issue #766, opened 2024-06-04
[BUG] Last two asserts of FuzzyGazeInteractorTestsForControllerlessRig::EyeGazeWorksAsExpectedTest() succeed in Editor but fail in pipeline.


ms-RistoRK:
## Describe the bug

Last two asserts of FuzzyGazeInteractorTestsForControllerlessRig::EyeGazeWorksAsExpectedTest() succeed in Editor but fail in pipeline.

## To reproduce

1. Uncomment assert lines.
2. Run locally in Editor
3. Run in pipeline
4. Compare results

## Expected behavior

Both pipeline and Editor test assert should throw the same result.

## Your setup (please complete the following information)

- Unity Version 2022.3.7f1
- MRTK XRI3 feature branch

## Target platform (please complete the following information)

- Build pipeline
- Editor

AMollis:
should be fixed with

- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/862

PR #862, opened 2024-07-17
Fix hands on quest 3 and fix eye-gaze fallbacks

AMollis:
# Overview

## Hand Tracking on Quest

Hand pose and selection was not working on Quest 3 when the "Meta Quest Touch Pro Controller Profile" was added a long side the Microsoft MR OpenXR Plugin "Hand Interaction" feature group.   This is because the rotation and position action were bound to the "meta quest controller" even though the controllers weren't active.  Adding checks for an "activeControler" fixes this.

Also, if the `HandPoseDriver` fallbacks to using joint data directly, the tracking state needed to be updated accordingly.  

For the pinch select input reader, the only stable way of determining if a pinch should be driven the raw joint data is by directly querying the HandPoseDriver.  The PinchInputReader now asks the HandPoseDrive if it is using raw joint data for poses.
## Gaze Fallbacks

Also, while fixing this on Quest, I noticed the head gaze fallback code wasn't working.   There were two places with this fallback.  A simple one in the "palm up solver", and the more generic one in TrackedPoseDriverWithFallback.   

The `TrackedPoseDriverWithFallback` wasn't fallbacking to head gaze because of some misplaced NOTs and ORs.  I fixed this and cleaned up the code a bit.   

The "palm up solver" wasn't computing tracking state like general "tracked pose driver", so this could lead to bugs (or inconsistencies).  Also, the "palm up solver" wasn't using the proper transform.

I also fixed the eye gaze tests.  There was a test that was made to test the incorrect behavior, I updated this.  Then there were tests that were commented out because they were failing due to buggy code.  I've reenabled these tests.

## Validated

- HoloLens 2 (with and without calibrated eyes)
- Quest 3 (with and without controllers)

## FIxes

- https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/766

### DIFF
--- a/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
+++ b/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
@@ -37,7 +37,7 @@ namespace MixedReality.Toolkit.Input
         /// <summary>
         /// Expose the tracking state for the hand pose driver, to allow <see cref="TrackedPoseDriverExtensions"/> to query it.
         /// </summary>
-        /// <remarks
+        /// <remarks>
         /// Avoid exposing this publicly as this <see cref="HandPoseDriver"/> is a workaround solution to support hand tracking on devices without interaction profiles.
         /// </remarks>
         internal InputTrackingState CachedTrackingState => m_trackingState;
--- a/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
+++ b/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
@@ -42,6 +42,11 @@ namespace MixedReality.Toolkit.Input
         /// </remarks>
         internal InputTrackingState CachedTrackingState => m_trackingState;
 
+        /// <summary>
+        /// Get if the last pose set was from a polyfill device pose. That is, if the last pose originated from the <see cref="XRSubsystemHelpers.HandsAggregator "/>.
+        /// </summary>
+        internal bool IsPolyfillDevicePose { get; private set; }
+
         #region Serialized Fields
         [Header("Hand Pose Driver Settings")]
 
@@ -74,7 +79,8 @@ namespace MixedReality.Toolkit.Input
             //
             // Note, for this workaround we need to consider the fact that the positon and rotation can be bound
             // to a control, but the control may not be active even if the tracking state is valid. So we need to
-            // check if there's an active control before using the position and rotation values.
+            // check if there's an active control before using the position and rotation values. If there's no active
+            // this means the action was not updated this frame and we should use the polyfill pose.
 
             bool missingPositionController =
                 (trackingType == TrackingType.RotationAndPosition || trackingType == TrackingType.PositionOnly) &&
@@ -89,7 +95,12 @@ namespace MixedReality.Toolkit.Input
                 TryGetPolyfillDevicePose(out Pose devicePose))
             {
                 m_trackingState = InputTrackingState.Position | InputTrackingState.Rotation;
+                IsPolyfillDevicePose = true;
                 ForceSetLocalTransform(devicePose.position, devicePose.rotation);
+            }
+            else
+            {
+                IsPolyfillDevicePose = false;
             }
         }
         #endregion TrackedPoseDriver Overrides
--- a/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
+++ b/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
@@ -85,7 +85,7 @@ namespace MixedReality.Toolkit.Input
                 (rotationInput.action == null || !rotationInput.action.HasAnyControls() || rotationInput.action.activeControl == null);
 
             // We will also check the tracking state here to account for a bound action but untracked interaction profile.
-            if ((missingPositionController || missingRotationController || m_trackingState == InputTrackingState.None) &&
+            if ((missingPositionController || missingRotationController || IsTrackingNone()) &&
                 TryGetPolyfillDevicePose(out Pose devicePose))
             {
                 m_trackingState = InputTrackingState.Position | InputTrackingState.Rotation;
--- a/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
+++ b/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
@@ -208,7 +208,7 @@ namespace MixedReality.Toolkit.Input
             // `TrackedPoseDriver` also sets the tracking state in a similar manner. Please see 
             // `TrackedPoseDriver::ReadTrackingState`. Replicating this logic in a subclass is not ideal, but it is
             // necessary since the base class does not expose the tracking state logic.
-            m_trackingState = this.GetInputTrackingStateNoCache();
+            m_trackingState = this.GetInputTrackingStateNoCache(handNode);
         }
 
         /// <summary>
--- a/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
+++ b/org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
@@ -55,7 +55,7 @@ namespace MixedReality.Toolkit.Input
         public XRNode HandNode => handNode;
         #endregion Serialized Fields
 
-        #region TrackedPoseDriver Overrides 
+        #region TrackedPoseDriver Overrides
         /// <inheritdoc />
         protected override void PerformUpdate()
         {
@@ -71,16 +71,24 @@ namespace MixedReality.Toolkit.Input
             // In case the pose input actions are not provided or not bound to a control, we will try to query the 
             // `HandsAggregator` subsystem for the device's pose. This logic and class should be removed once we 
             // have universal hand interaction profile(s) across vendors.
-            bool missingPositionController = (trackingType.HasFlag(TrackingType.PositionOnly) || trackingType.HasFlag(TrackingType.RotationAndPosition)) &&
-                (positionInput.action == null || !positionInput.action.HasAnyControls());
+            //
+            // Note, for this workaround we need to consider the fact that the positon and rotation can be bound
+            // to a control, but the control may not be active even if the tracking state is valid. So we need to
+            // check if there's an active control before using the position and rotation values.
 
-            bool missingRotationController = (trackingType.HasFlag(TrackingType.RotationOnly) || trackingType.HasFlag(TrackingType.RotationAndPosition)) &&
-                (rotationInput.action == null || !rotationInput.action.HasAnyControls());
+            bool missingPositionController =
+                (trackingType == TrackingType.RotationAndPosition || trackingType == TrackingType.PositionOnly) &&
+                (positionInput.action == null || !positionInput.action.HasAnyControls() || positionInput.action.activeControl == null);
+
+            bool missingRotationController =
+                (trackingType == TrackingType.RotationAndPosition || trackingType == TrackingType.RotationOnly) &&
+                (rotationInput.action == null || !rotationInput.action.HasAnyControls() || rotationInput.action.activeControl == null);
 
             // We will also check the tracking state here to account for a bound action but untracked interaction profile.
-            if ((missingPositionController || missingRotationController || IsTrackingNone()) &&
+            if ((missingPositionController || missingRotationController || m_trackingState == InputTrackingState.None) &&
                 TryGetPolyfillDevicePose(out Pose devicePose))
             {
+                m_trackingState = InputTrackingState.Position | InputTrackingState.Rotation;
                 ForceSetLocalTransform(devicePose.position, devicePose.rotation);
             }
         }
@@ -208,7 +216,7 @@ namespace MixedReality.Toolkit.Input
             // `TrackedPoseDriver` also sets the tracking state in a similar manner. Please see 
             // `TrackedPoseDriver::ReadTrackingState`. Replicating this logic in a subclass is not ideal, but it is
             // necessary since the base class does not expose the tracking state logic.
-            m_trackingState = this.GetInputTrackingStateNoCache(handNode);
+            m_trackingState = this.GetInputTrackingStateNoCache();
         }
 
         /// <summary>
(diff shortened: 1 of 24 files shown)


AMollis:
If anyone is able to test on Quest 3 today/tomorrow, let me know.  Otherwise I will test Friday

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverExtensions.cs
```diff
@@ -84,14 +89,27 @@ internal static InputTrackingState GetInputTrackingStateNoCache(this TrackedPose
...
             if (trackingStateAction.HasAnyControls())
             {
-                return (InputTrackingState)trackingStateAction.ReadValue<int>();
+                result = (InputTrackingState)trackingStateAction.ReadValue<int>();
             }
             else
             {
-                return InputTrackingState.None;
+                result = InputTrackingState.None;
             }
```
keveleigh, line 100: nit: this `else` could be removed now that `result` is already initialized to `InputTrackingState.None`

keveleigh (commented):
No Quest to help test, unfortunately

MaxPalmer-UH:
I'll try to test this today.

MaxPalmer-UH:
With this branch I see rays when using controllers appear above / below the horizon, as expected. I'd like to double check which options should be on for hand trackling + MRTK + Quest3 though? Strangly, if I enable hand tracking in the OpenXR list (which I assume should be off), I see the issue with rays disappaering when angled below the 'horizon'.

image

Which options do you believe should be enabled for Meta Quest hand tracking on Quest 3 with MRTK3?

When I drop the controllers and it switches to hands, I see no hand rays at all.

MaxPalmer-UH:
With that option on, controllers and tracked hands both show rays that diappear below the horizon.
With than option off, controller show rays all the time, but hands show no rays.
That's with your branch/fork @AMollis 
Also, can I check which subsystems should be enabled under MRTK too please? This is what I'm using:
**
image
**

org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
```diff
@@ -354,6 +356,39 @@ public void StartWorldLockReattachCheckCoroutine()
...
             new ProfilerMarker("[MRTK] HandConstraintPalmUp.TryGenerateHandPlaneAndActivationPoint");
 
+        /// <summary>
+        /// Get the input tracking state.
+        /// </summary>
+        private InputTrackingState GetInputTrackingState(TrackedPoseDriver driver)
+        {
+            // Note, that the logic in this class is meant to reproduce the same logic as the base. The base
+            // `TrackedPoseDriver` also sets the tracking state in a similar manner. Please see 
+            // `TrackedPoseDriver::ReadTrackingState`. Replicating this logic in a subclass is not ideal, but it is
```
AMollis, line 366: I should update this comment

org.mixedrealitytoolkit.input/Tracking/HandPoseDriver.cs
```diff
@@ -71,16 +71,24 @@ protected override void PerformUpdate()
...
+                (trackingType == TrackingType.RotationAndPosition || trackingType == TrackingType.PositionOnly) &&
+                (positionInput.action == null || !positionInput.action.HasAnyControls() || positionInput.action.activeControl == null);
+
+            bool missingRotationController =
+                (trackingType == TrackingType.RotationAndPosition || trackingType == TrackingType.RotationOnly) &&
+                (rotationInput.action == null || !rotationInput.action.HasAnyControls() || rotationInput.action.activeControl == null);
 
             // We will also check the tracking state here to account for a bound action but untracked interaction profile.
-            if ((missingPositionController || missingRotationController || IsTrackingNone()) &&
+            if ((missingPositionController || missingRotationController || m_trackingState == InputTrackingState.None) &&
```
AMollis, line 88: should un do this change

AMollis:
> With that option on, both controllers and tracked hands show rays that disappear below an arbitrary horizon. With than option off, controller show rays all the time, but hands show no rays. That's with your branch/fork @AMollis on a Quest3. Also, can I check which subsystems should be enabled under MRTK too please? This is what I'm using: ** image **

Yeah my fix was WAY OFF...  I got my hands on a Quest, and have made a real fix :)

MaxPalmer-UH:
> > With that option on, both controllers and tracked hands show rays that disappear below an arbitrary horizon. With than option off, controller show rays all the time, but hands show no rays. That's with your branch/fork @AMollis on a Quest3. Also, can I check which subsystems should be enabled under MRTK too please? This is what I'm using: ** image **
> 
> Yeah my fix was WAY OFF... I got my hands on a Quest, and have made a real fix :)

Yes, can confirm, now working for me too. Thanks.

shaynie (approved):

org.mixedrealitytoolkit.input/Tracking/TrackedPoseDriverWithFallback.cs
```diff
@@ -71,41 +69,59 @@ protected override void PerformUpdate()
...
+                hasRotationFallbackAction &&
+                fallbackInputTrackingState.HasFlag(InputTrackingState.Rotation))
             {
-                neededToGetFallbackData = true;
+                fallbackDataUsed |= InputTrackingState.Rotation;
                 rotation = fallbackRotationAction.action.ReadValue<Quaternion>();
             }
 
-            if (neededToGetFallbackData) //because either position, rotation, or both data were obtained from fallback actions
+            // Id either position, rotation, or both data were obtained from fallback actions,
```
ms-RistoRK, line 115:
```suggestion
            // If either position, rotation, or both data were obtained from fallback actions,
```

ms-RistoRK (dismissed):

org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
```diff
@@ -294,16 +295,17 @@ private bool IsUserGazeMeetingThresholdRequirements(XRNode hand)
                     }
                 }
                 #pragma warning restore CS0618
-                else if (TrackedPoseDriverLookup != null)
+                else if (TrackedPoseDriverLookup != null &&
+                    TrackedPoseDriverLookup.GazeTrackedPoseDriver != null)
```
whebertML, line 299: Putting the check for GazeTrackedPoseDriver != null here, causes the logic to skip the fallback to headpose using camera transform.

Perhaps include a null check of the driver inside GetInputTrackingState, where it just returns InputTrackingState.None in that case, allowing the fallback.

whebertML:
@AMollis   Testing with Quest 3.

Seeing an issue of no fallback for HandMenu if the Gaze Controller TrackedPoseDriver isn't defined in TrackedPoseDriverLookup.  That should fallback to using headpose via camera transform.

Also, I don't appear to be getting Select values when switching to the Quest 3 controller in the new rig.  The legacy rig still works fine when using controller, but I don't see select value change when pulling trigger in new rig.  Here is my OpenXR config for reference - was seeing this with either the "Meta Quest Touch Pro Controller Profile" or the "Oculus Touch Controller Profile".

AMollis:
> @AMollis Testing with Quest 3.
> 
> Seeing an issue of no fallback for HandMenu if the Gaze Controller TrackedPoseDriver isn't defined in TrackedPoseDriverLookup. That should fallback to using headpose via camera transform.
> 
> Also, I don't appear to be getting Select values when switching to the Quest 3 controller in the new rig. The legacy rig still works fine when using controller, but I don't see select value change when pulling trigger in new rig. Here is my OpenXR config for reference - was seeing this with either the "Meta Quest Touch Pro Controller Profile" or the "Oculus Touch Controller Profile".
> 
> image

@whebertML  your settings should look like this.  I don't think Query 3 supports the EXT_Hand_Interaction.  So fallbacks using the joint data from Unity's "Hand Tracking Subsystem" feature, Controller data from our "Hand Tracking" feature, Controller models from our "Motion Controller Model"

image

And I find better joint positioning when using Unity's Hand Package

image


Also....the default MRTK3 input actions don't have bindings for those other "Meta" features....so not entirely sure what those will do.

whebertML:
@AMollis 

Interesting, I enabled some of the features like you had, but no difference for me - still no controller selection, though it does track.  Again, the controller totally works with the legacy rig, but for some reason not the new rig, for me.  So you are seeing controller selection working in the new rig when switching from hands to controller?

image

I wonder if it has something to do with using a different meta OpenXR package.  I notice your Meta feature list is a little different than mine.  I'm using the com.unity.xr.meta-openxr package, version 1.0.1, which is what I thought we were supposed to use for Quest 3, for features like passthrough and such.  What package are you using?

image

But, it is interesting that the Quest 3 controller works in the legacy rig and I can switch between hand and controller no issue.  But in the new rig, with this change I think, the controller selection stops working for me (with the above config).  I'll go back and test if that is also the case before this change.

org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
AMollis: Ahh yes :)  I made this better.   See the new helper function for details

AMollis:
> @AMollis
> 
> Interesting, I enabled some of the features like you had, but no difference for me - still no controller selection, though it does track. Again, the controller totally works with the legacy rig, but for some reason not the new rig, for me. So you are seeing controller selection working in the new rig when switching from hands to controller?
> 
> image
> 
> I wonder if it has something to do with using a different meta OpenXR package. I notice your Meta feature list is a little different than mine. I'm using the com.unity.xr.meta-openxr package, version 1.0.1, which is what I thought we were supposed to use for Quest 3, for features like passthrough and such. What package are you using?
> 
> image
> 
> But, it is interesting that the Quest 3 controller works in the legacy rig and I can switch between hand and controller no issue. But in the new rig, with this change I think, the controller selection stops working for me (with the above config). I'll go back and test if that is also the case before this change.

@whebertML .... I was using the All-In-One package from Meta's publication on the Asset Store...per their documentation

https://assetstore.unity.com/packages/tools/integration/meta-xr-all-in-one-sdk-269657

whebertML:
@AMollis That looks like a much more up to date package (June of 2024).  I searched some time ago and saw this package, https://docs.unity3d.com/Packages/com.unity.xr.meta-openxr@1.0/manual/index.html, and another document referencing it, took it for the latest info from Meta.  Guess not.  Thanks, will switch to using the latest one and see if it changes (I assume it will).

AMollis:
> > @AMollis
> > Interesting, I enabled some of the features like you had, but no difference for me - still no controller selection, though it does track. Again, the controller totally works with the legacy rig, but for some reason not the new rig, for me. So you are seeing controller selection working in the new rig when switching from hands to controller?
> > image
> > I wonder if it has something to do with using a different meta OpenXR package. I notice your Meta feature list is a little different than mine. I'm using the com.unity.xr.meta-openxr package, version 1.0.1, which is what I thought we were supposed to use for Quest 3, for features like passthrough and such. What package are you using?
> > image
> > But, it is interesting that the Quest 3 controller works in the legacy rig and I can switch between hand and controller no issue. But in the new rig, with this change I think, the controller selection stops working for me (with the above config). I'll go back and test if that is also the case before this change.
> 
> @whebertML .... I was using the All-In-One package from Meta's publication on the Asset Store...per their documentation
> 
> https://assetstore.unity.com/packages/tools/integration/meta-xr-all-in-one-sdk-269657

I'll add in Untiy's [com.unity.xr.meta-openxr](https://docs.unity3d.com/Packages/com.unity.xr.meta-openxr@1.0/manual/index.html) and see what happens

keveleigh:
> I was using the All-In-One package from Meta's publication on the Asset Store...per their documentation

For base support, we definitely _shouldn't_ need this (or even the Meta OpenXR package). In the past, all we needed was the `Meta Quest Support` feature and the `Oculus Touch Controller Profile` from Unity's base OpenXR Plugin package.

whebertML:
Trying the all-in-one Meta package, I have other issues.  Like Hand selection not working in the legacy rig, though controller selection does.  But in the new rig, hands work, but the controller selection is "sticky", and hard to release.  Could be something odd with going back-and-forth in configuration.  Might need to start over clean - but may have to wait till next week for me.

AMollis:
> t the controller selection is "sticky", an

Yes,... the sticky thing this is known.  I'll tackle that in a different PR.

AMollis:
> > t the controller selection is "sticky", an
> 
> Yes,... the sticky thing this is known. I'll tackle that in a different PR.

This change at least get's quest 3 mostly working.   

We should get this change in and investigate other failures with additional PRs

AMollis:
> > > t the controller selection is "sticky", an
> > 
> > 
> > Yes,... the sticky thing this is known. I'll tackle that in a different PR.
> 
> This change at least get's quest 3 mostly working with new rig. Minus the sticky behavior we're good.
> 
> We should get this change in and investigate other failures with additional PRs.
> 
> @whebertML is there something in particular you'd like to see changed with this PR? Happy to investigate other failures in a future PR. But this PR should be introducing any new regressions.

@whebertML  thanks for all the feedback :) 

I found the issue with the sticky select with Quest 3 controls.  It was a fatal flaw in my fix to the "PinchInputReader".  I can't just rely on selectInput.action.activeControl == null to determine if pitch should be driven by joint poses.   For actions action.activeControl is only non-null if the state was updated in the current frame. For buttons this only happens when you are actively pressing the button.  

As a result, the only stable way of determining if pinch should be using joint data is by querying the HandPoseDriver, and asking the HandPoseDriver if it is also using the joint poses directly (ie isPolyfilledPose).

This means I have to pass in the HandPoseDriver to the PinchInputReader...not horrible :)  And results in less code in the PinchInputReader.

shaynie (dismissed):

AMollis:
@whebertML ... please take a look at these latest changes.   I plan on merge this early tomorrow, to unblock the preview release.

shaynie (approved):

AMollis:
> @whebertML ... please take a look at these latest changes. I plan on merging this early tomorrow, to unblock the preview release.

@whebertML  I'm going to merge this to unblock preview release.  I'll address additional feedback with future PR

whebertML:
@AMollis Testing latest, but starting off with a clean slate with just latest feature/XRI3 (not yet pulling this PR) and no special Quest packages - just the base Meta Quest Support and Oculus Touch Controller Profile.  With that, I seem to have hands, hand rays, and Controllers working fine with both the legacy and new XRI3 rigs.

As @keveleigh suggests, seems things are fine at least with the base support.  So does that mean adding specific meta packages may, or just using specific profiles, has issues?  Will try a couple of combos.  Just wanting to be sure what issues the changes are actually intended to fix.

PR #864, opened 2024-07-18
Fix issue with the HandMenu not working in new controllerless rig

whebertML:
The HandMenu was not working in the new controllerless rig.  This is a simple fix to use the GazeTrackedPoseDriver's transform, but also makes use of the TrackedPoseDriver extension TryGetTrackingState(...), which required an assembly reference on Input. 

Does this also mean SpatialManipulation should have a hard dependency on Input in package.manifest?

Solver.cs was already using TrackedPoseDriverLookup, defined in Input, so wouldn't that already cause dependency?

image

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
@@ -6,6 +6,7 @@ using Unity.Profiling;
 using UnityEngine;
 using UnityEngine.Serialization;
 using UnityEngine.XR;
+using MixedReality.Toolkit.Input;
 
 namespace MixedReality.Toolkit.SpatialManipulation
 {
@@ -296,14 +297,14 @@ namespace MixedReality.Toolkit.SpatialManipulation
                 #pragma warning restore CS0618
                 else if (TrackedPoseDriverLookup != null)
                 {
-                    InputTrackingState gazeTrackingStateInput = (InputTrackingState)TrackedPoseDriverLookup.GazeTrackedPoseDriver.trackingStateInput.action.ReadValue<int>();
                     if (TrackedPoseDriverLookup.GazeTrackedPoseDriver != null &&
+                        TrackedPoseDriverLookup.GazeTrackedPoseDriver.TryGetTrackingState(out InputTrackingState gazeTrackingStateInput) &&
                         gazeTrackingStateInput.HasFlag(InputTrackingState.Position) &&
                         gazeTrackingStateInput.HasFlag(InputTrackingState.Rotation))
                     {
                         gazeRay = new Ray(
-                                TrackedPoseDriverLookup.transform.position,
-                                TrackedPoseDriverLookup.transform.forward);
+                                TrackedPoseDriverLookup.GazeTrackedPoseDriver.transform.position,
+                                TrackedPoseDriverLookup.GazeTrackedPoseDriver.transform.forward);
                         usedEyeGaze = true;
                     }
                     else
--- a/org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
@@ -297,7 +297,7 @@ namespace MixedReality.Toolkit.SpatialManipulation
                 else if (TrackedPoseDriverLookup != null)
                 {
                     InputTrackingState gazeTrackingStateInput = TrackedPoseDriverLookup.GazeTrackedPoseDriver != null ?
-                        (InputTrackingState)TrackedPoseDriverLookup.GazeTrackedPoseDriver.trackingStateInput.action?.ReadValue<int>() :
+                        (InputTrackingState)(TrackedPoseDriverLookup.GazeTrackedPoseDriver.trackingStateInput.action?.ReadValue<int>() ?? default) :
                         InputTrackingState.None;
 
                     if (gazeTrackingStateInput.HasFlag(InputTrackingState.Position) &&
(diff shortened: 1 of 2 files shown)


org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
```diff
@@ -296,14 +297,14 @@ private bool IsUserGazeMeetingThresholdRequirements(XRNode hand)
                 #pragma warning restore CS0618
                 else if (TrackedPoseDriverLookup != null)
                 {
-                    InputTrackingState gazeTrackingStateInput = (InputTrackingState)TrackedPoseDriverLookup.GazeTrackedPoseDriver.trackingStateInput.action.ReadValue<int>();
```
whebertML, line 299: This line could potentially cause a NullRefException if GazeTrackedPoseDriver isn't defined or the action isn't defined, so opted for the more safe TryGetTrackingState below.

keveleigh:
> Does this also mean SpatialManipulation should have a hard dependency on Input in package.manifest?

Not introducing a new package dependency would be ideal! Should the extensions live in Core instead, if they're used across multiple packages?

whebertML:
> > Does this also mean SpatialManipulation should have a hard dependency on Input in package.manifest?
> 
> Not introducing a new package dependency would be ideal! Should the extensions live in Core instead, if they're used across multiple packages?

@keveleigh Though Solver.cs already utilizes TrackedPoseDriverLoopup, which is defined in Input.  But perhaps TrackedPoseDriverLookup should be moved to Core along with the Extension.  The previous, now deprecated, ControllerLookup is in Core, so that makes sense.

keveleigh:
> UPDATE: Actually, I was wrong. TrackedPoseDriverLookup is in Core - but lives in Input namespace (which is why I thought it was in Input). So should the namespace just be corrected and the TrackedPoseDriverExtension be moved?

Yeah, that sounds right to me!

whebertML:
I see why TrackedPoseDriverExtensions doesn't live in Core, because it references HandPoseDriver, which lives in Input, so can't have the cyclic dependency.

whebertML:
@keveleigh Simplified to avoid the dependency on assembly.

org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
whebertML: No longer using TryGetTrackingState, just modified existing logic slightly to avoid possible NullRefException instead.

org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
```diff
@@ -296,14 +296,16 @@ private bool IsUserGazeMeetingThresholdRequirements(XRNode hand)
...
-                    if (TrackedPoseDriverLookup.GazeTrackedPoseDriver != null &&
-                        gazeTrackingStateInput.HasFlag(InputTrackingState.Position) &&
+                    InputTrackingState gazeTrackingStateInput = TrackedPoseDriverLookup.GazeTrackedPoseDriver != null ?
+                        (InputTrackingState)(TrackedPoseDriverLookup.GazeTrackedPoseDriver.trackingStateInput.action?.ReadValue<int>() ?? default) :
```
keveleigh, line 300: Do we want to explicitly set this to `None` in the null case?

```suggestion
                        (InputTrackingState)(TrackedPoseDriverLookup.GazeTrackedPoseDriver.trackingStateInput.action?.ReadValue<int>() ?? InputTrackingState.None) :
```

keveleigh: Ohh or does this run into that not-yet-available-in-Unity feature where each side of the `??` is returning different types?

keveleigh (approved):

AMollis:
> > Does this also mean SpatialManipulation should have a hard dependency on Input in package.manifest?
> 
> Not introducing a new package dependency would be ideal! Should the extensions live in Core instead, if they're used across multiple packages?

Aggreed

AMollis (changes requested):
I also fixed this...in a somewhat similar way.  However, I wanted to calculate "tracking state" the same way the tracked pose driver does.

Please see my PR

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/862/files#diff-689651bf85edb79c910ea2d6a9f740017037a524d5c9ab7eb77cf26bf8ecfa6c

org.mixedrealitytoolkit.spatialmanipulation/Solvers/HandConstraintPalmUp.cs
whebertML: Yeah, it was complaining if I tried that explicit value, but default should just be a 0 int, which is effectively the same thing and gets cast as None.

Adam has fixed this section of code in his larger PR, so this likely won't be used anyway.

AMollis:
@whebertML  how do you want to handle this....Happy to merge in your changes.   But I would like the tracking state logic to match the tracked pose driver.   I can do this with my PR, or you can with your PR.

whebertML:
Hey @AMollis , sorry, been preoccupied with a task.  I think the change can just come from your PR.  Though I did want to test something with that:

If the gaze tracked pose driver is considered tracking even without a reference or bound control, won't that end up with Identity for a transform being used instead of the fallback to headpose?

AMollis:
> Hey @AMollis , sorry, been preoccupied with a task. I think the change can just come from your PR. Though I did want to test something with that:
> 
> If the gaze tracked pose driver is considered tracking even without a reference or bound control, won't that end up with Identity for a transform being used instead of the fallback to headpose?

Unity's TrackedPoseDriver behaves this way.   The assumption in TrackedPoseDriver, is that if there is no "tracking binding" assume the controller pose is valid.

I would like the same behavior in this class as well, to avoid diverging behaviors in the GazeInteractor Pose and the PalmUp solver

whebertML:
Was preoccupied with an internal task for a while, now testing your PR on Quest3.

AMollis (approved):

PR #868, opened 2024-07-18
XRI3 migration bump xr interaction toolkit package version to 3.0.4

ms-RistoRK:


AMollis:
With some future change, we should also add in support for the [EXT_Hand_Interaction](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XR_EXT_hand_interaction) profile.  XRI 3.0.4 supports this now.

We'd need to update our default MRTK actions to include EXT_Hand_Interaction bindings.   And also update the sample project to add the EXT_Hand_Interaction profile.

https://docs.unity3d.com/Packages/com.unity.xr.openxr@1.10/manual/features/handinteractionprofile.html

AMollis (approved):

Issue #671, opened 2024-03-08
[BUG] SpatialManipulationReticle throws NullReferenceException


anonymous2585:
## Describe the bug

The function `RotateReticle` of `SpatialManipulationReticle` access a transform parent. In some conditions, this parent can be null, causing a NullReferenceException.
The transform given to this function is wrong, instead of the handle (which should always have a parent), the main object transform is given because [it's always the hover index 0 that is given to the function](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/c3af2c27de98c0caa96ce0be076de6127b526139/org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SpatialManipulationReticle.cs#L50) (see video).

## To reproduce

1. Open the scene `BoundsControlExamples`
2. On the right hand `Far Ray`, uncheck `Hit closest only`
3. Place the object `CoffeeBoundsControl (1)` at the root of the scene
4. On the object `CoffeeBoundsControl (1)` component `BoundsControl` check `Handles active`
5. Press play
6. Activate the hand and target to a handle that is on the back face of the cube
7. See the exception

```
NullReferenceException: Object reference not set to an instance of an object
MixedReality.Toolkit.SpatialManipulation.SpatialManipulationReticle.GetCursorTargetAxes (UnityEngine.Vector3 normal, UnityEngine.Vector3& right, UnityEngine.Vector3& up, UnityEngine.Vector3& forward, UnityEngine.Transform contextTransform) (at C:/Users/xxx/MixedRealityToolkit-Unity/org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SpatialManipulationReticle.cs:127)
MixedReality.Toolkit.SpatialManipulation.SpatialManipulationReticle.RotateReticle (UnityEngine.Vector3 reticleNormal, UnityEngine.Transform hitTargetTransform) (at C:/Users/xxx/MixedRealityToolkit-Unity/org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SpatialManipulationReticle.cs:73)
[...]

PR #873, opened 2024-07-24
Fix null ref in SpatialManipulationReticle when multiple interactables are hovered

keveleigh:
Fixes https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/671

Ensures that we actually pass a `BoundsHandleInteractable` to `RotateReticle`, as that's what it's expecting as far as hierarchy. If we inadvertently pass the `BoundsControl`, the hierarchy logic is incorrect (no parent when it's in the root of the scene).

This case won't be hit in the default MRTK XR Rig. It's only when "Hit Closest Only" on a ray interactor is unchecked that this case is possible.

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SpatialManipulationReticle.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SpatialManipulationReticle.cs
@@ -55,7 +55,7 @@ namespace MixedReality.Toolkit.SpatialManipulation
                     {
                         for (int i = 0; i < hoveredCount; i++)
                         {
-                            if (hoveredInteractables[i] is ISnapInteractable)
+                            if (hoveredInteractables[i] is BoundsHandleInteractable)
                             {
                                 interactableIndex = i;
                                 break;
--- a/org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SpatialManipulationReticle.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SpatialManipulationReticle.cs
@@ -2,6 +2,7 @@
 // Licensed under the BSD 3-Clause
 
 using System;
+using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.XR.Interaction.Toolkit;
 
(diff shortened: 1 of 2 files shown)


org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SpatialManipulationReticle.cs
```diff
@@ -47,7 +47,23 @@ public void UpdateVisual(ReticleVisualUpdateArgs args)
                 }
                 else if (rayInteractor.interactablesHovered.Count > 0)
                 {
-                    RotateReticle(args.ReticleNormal, rayInteractor.interactablesHovered[0].transform);
+                    int interactableIndex = 0;
+
+                    List<IXRHoverInteractable> hoveredInteractables = rayInteractor.interactablesHovered;
```
shaynie, line 52: Should cache this to avoid the per-frame allocation

shaynie (changes requested):

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/Visuals/SpatialManipulationReticle.cs
keveleigh: There's already no per-frame allocation, luckily! The `.interactablesHovered` [property's documentation](https://docs.unity3d.com/Packages/com.unity.xr.interaction.toolkit@3.0/api/UnityEngine.XR.Interaction.Toolkit.Interactors.IXRHoverInteractor.html#UnityEngine_XR_Interaction_Toolkit_Interactors_IXRHoverInteractor_interactablesHovered) states:

> It is exposed as a List<T> rather than an IReadOnlyList<T> to avoid GC Allocations when enumerating the list.

and it calls into [a Unity CoreUtils helper](https://docs.unity3d.com/Packages/com.unity.xr.core-utils@2.4/api/Unity.XR.CoreUtils.Collections.HashSetList-1.html#Unity_XR_CoreUtils_Collections_HashSetList_1_AsList) that itself says:

> Exposes internal list without any allocation.

I also went ahead and profiled to verify:
image

shaynie: Awesome, I learned something.

shaynie (approved):

AMollis (approved):

Issue #890, opened 2024-08-16
[XRI3] [BUG] Rigged hand mesh visualizer gets stuck "on" when using a hand interaction profile


keveleigh:
## Describe the bug

When using a hand interaction profile, the rigged hand mesh visualizer's select action reader logic causes the selection outline highlighting to be stuck on after an initial pinch.

## To reproduce

1. Enable the Hand Interaction Profile in the OpenXR settings (may also repro with the Microsoft Hand Interaction Profile)
2. Ensure `pointerActivateValue` is present in your select action mappings
    1. ![image](https://github.com/user-attachments/assets/3cf1fa67-502c-4f99-a8d6-f620c6c4dc93)
2. Run an MRTK examples with the rigged hand mesh (like HandInteractionExamples)
3. Pinch!

## Expected behavior

On unpinch, the outline highlighting will return to the default, unhighlighted state

## Screenshots

![image](https://github.com/user-attachments/assets/ea69e3f2-bbe9-45af-adb5-c57b51f00415)

shaynie:
Fixed with #891

PR #891, opened 2024-08-16
Fix rigged hand mesh with input actions

keveleigh:
Fixes https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/890

Updates `RiggedHandMeshVisualizer` to treat a non-performed action as a 0 "selectedness".

> [!IMPORTANT]  
>
>I think we may want to consider reverting https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/852 (pinging @whebertML for his thoughts). The default implementation of `XRInputButtonReader` _does_ only return `true` when the action is in progress:
>
>image
>
>I think the underlying issue was that we had at least one code path (the change in this PR) that wasn't properly treating a return value of `false` as equivalent to `0`. Where you seeing the previous return value be problematic in other places too?

### DIFF
--- a/org.mixedrealitytoolkit.input/Controllers/HandModel.cs
+++ b/org.mixedrealitytoolkit.input/Controllers/HandModel.cs
@@ -109,8 +109,12 @@ namespace MixedReality.Toolkit.Input
             {
                 modelParent = new GameObject($"[{gameObject.name}] Model Parent").transform;
                 modelParent.SetParent(transform, false);
+#if HAS_SET_LOCAL_POSITION_AND_ROTATION
+                modelParent.SetLocalPositionAndRotation(Vector3.zero, Quaternion.identity);
+#else
                 modelParent.localPosition = Vector3.zero;
                 modelParent.localRotation = Quaternion.identity;
+#endif
             }
         }
     }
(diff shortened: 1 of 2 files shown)


AMollis (approved):

whebertML (approved):
@keveleigh ,  regarding #852:

I believe the changes in #852 actually made it reflect the same behavior as that of `IsInProgress()` as that returns true during the entire range of select values, whereas before it was only returning true when isPinched was finally reached (what `isPerformed` represented in this fallback logic) - skipping the ranges from > 0 and < 1.

This change looks good to me.  👍

keveleigh:
> @keveleigh , regarding #852:
> 
> I believe the changes in #852 actually made it reflect the same behavior as that of `IsInProgress()` as that returns true during the entire range of select values, whereas before it was only returning true when isPinched was finally reached (what `isPerformed` represented in this fallback logic) - skipping the ranges from > 0 and < 1.
> 
> This change looks good to me. 👍

@whebertML Ohh interesting! I may have been misunderstanding Unity's docs then. I'll have to test on-device to verify.

From their docs, and action goes into a `Canceled` state (and thus `IsInProgress() == false`) when:

> By default, a Button action cancels when a control falls back below the button press threshold (see defaultButtonPressPoint) and a Value action cancels when a control moves back to its default value.

which reads to me as "when it returns to `false` or `0`".

Issue #885, opened 2024-08-06
[XRI3] [BUG] PinchInputReader can get stuck using polyfill pinch detection


whebertML:
## Describe the bug

PinchInputReader can get stuck using polyfill pinch detection, even when the user has switched from hands to controller that has an input profile that has bound controls on the input actions.

## To reproduce

1. Configure the MRTK3, version 4.0.0-pre.1, project to build for Quest, and build+deploy the app to a Meta Quest 2 or 3.
2. While using hands (not controller), start the app, and continue using only hands when the app comes up.
3. Confirm pinch selection using hands, which on the Quest 2/3 should be utilizing polyfill.
4. Switch to using the Quest Controller, and notice pinch selection with the controller trigger does not work.
5. Note: If the app is started using the controller, pinch selection works with the controller, and the user can switch between hands and controller.  It is only if the app is started with hands where the issue arises. 

This is due to the logic to cache the input action validity for the select & select values upon startup getting locked to false when using hands, and these flags never get re-evaluated when switching input from hands to controller.  When they are false, the logic always goes to polyfill.  When they are true, as when starting the app with the controller, the logic ends up switching between polyfill and the input actions based on GetIsPolyfillDevicePose().  See code below:

```
            if (!m_isSelectionActionValidCache || !m_isSelectionActionValueValidCache || GetIsPolyfillDevicePose())
            {
                UpdatePinchSelection();
            }
```

Should we only rely on GetIsPolyfillDevicePose(), or perhaps call UpdateActionValidCaches() every frame instead of only first frame?

## Expected behavior

Pinch selection works on the Quest Controller after having switched from hands to controller.

## Your setup (please complete the following information)

- Unity Version [2022.3.27f1]
- MRTK Version or Commit [v4.0.0-pre.1]

## Target platform (please complete the following information)

- Quest 2 or 3

whebertML:
@AMollis FYI ^ 
Discovered the issue I was seeing on Quest with the controller and pinch select.

AMollis:
Fixed this #892

PR #704, opened 2024-03-25
Add logic to account for a bound but untracked interaction profile

keveleigh:
This could show up on runtimes where a controller is disconnected, hand tracking spins up, but the interaction profile is not cleared. This is allowed, per-spec: "The runtime may return the last-known interaction profile in the event that no controllers are active."

If we have a bound but untracked interaction profile, we now make an attempt at polyfilling based on hand tracking data.

Fixes #703

PR #892, opened 2024-08-16
Check input action availability every frame instead of once on startup

keveleigh:
Fixes https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/885

Since we were caching the input action availability once on startup, it became impossible to switch between input actions and the polyfill fallback, such as when switching between hands and controllers on a device that doesn't support a hand interaction profile or in an app that doesn't have a hand interaction profile enabled.

This PR updates the input reader logic to check the input action availability once each frame, caching that value for all subsequent uses in the same frame. Since this MonoBehaviour runs at position `k_XRInputDeviceButtonReader = -31000`, it's basically the first thing that runs and it's guaranteed to cache these values before any of it's `Read` methods are called.

### DIFF
--- a/org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
+++ b/org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
@@ -104,9 +104,9 @@ namespace MixedReality.Toolkit.Input
         #region Private Fields
 
         private FallbackState m_fallbackState;
-        private bool m_firstUpdate = true;
-        private bool m_isSelectionActionValidCache = false;
-        private bool m_isSelectionActionValueValidCache = false; 
+        private bool m_isTrackingStatePolyfilled = false;
+        private bool m_isSelectPolyfilled = false;
+        private bool m_isSelectValuePolyfilled = false;
 
         private static readonly ProfilerMarker UpdatePinchSelectionPerfMarker =
             new ProfilerMarker("[MRTK] PinchInputReader.UpdatePinchSelection");
@@ -122,10 +122,6 @@ namespace MixedReality.Toolkit.Input
         {
             selectAction.EnableDirectAction();
             selectActionValue.EnableDirectAction();
-
-            // Read current input values when becoming enabled,
-            // but wait until after the input update so the input is read at a consistent time
-            m_firstUpdate = true;
 
             // reset fallback state
             m_fallbackState = default;
@@ -145,16 +141,14 @@ namespace MixedReality.Toolkit.Input
         /// </summary>
         protected virtual void Update()
         {
-            if (m_firstUpdate)
-            {
-                UpdateActionValidCaches();
-                m_firstUpdate = false;
-            }
+            m_isTrackingStatePolyfilled = trackedPoseDriver.GetIsPolyfillDevicePose();
+            m_isSelectPolyfilled = !selectAction.action.HasAnyControls();
+            m_isSelectValuePolyfilled = !selectActionValue.action.HasAnyControls();
 
             // Workaround for missing select actions on devices without interaction profiles
             // for hands, such as Varjo and Quest. Should be removed once we have universal
             // hand interaction profile(s) across vendors.
-            if (!m_isSelectionActionValidCache || !m_isSelectionActionValueValidCache || GetIsPolyfillDevicePose())
+            if (m_isTrackingStatePolyfilled || m_isSelectPolyfilled || m_isSelectValuePolyfilled)
             {
                 UpdatePinchSelection();
             }
@@ -167,10 +161,10 @@ namespace MixedReality.Toolkit.Input
         /// <inheritdoc />
         public bool ReadIsPerformed()
         {
-            if (m_isSelectionActionValidCache && !GetIsPolyfillDevicePose())
+            if (!m_isSelectPolyfilled && !m_isTrackingStatePolyfilled)
             {
-                var action = selectAction.action;
-                var phase = action.phase;
+                InputAction action = selectAction.action;
+                InputActionPhase phase = action.phase;
                 return phase == InputActionPhase.Performed || (phase != InputActionPhase.Disabled && action.WasPerformedThisFrame());
             }
             else
@@ -182,7 +176,7 @@ namespace MixedReality.Toolkit.Input
         /// <inheritdoc />
         public bool ReadWasPerformedThisFrame()
         {
-            if (m_isSelectionActionValidCache && !GetIsPolyfillDevicePose())
+            if (!m_isSelectPolyfilled && !m_isTrackingStatePolyfilled)
             {
                 return selectAction.action.WasPerformedThisFrame();
             }
@@ -195,7 +189,7 @@ namespace MixedReality.Toolkit.Input
         /// <inheritdoc />
         public bool ReadWasCompletedThisFrame()
         {
-            if (m_isSelectionActionValidCache && !GetIsPolyfillDevicePose())
+            if (!m_isSelectPolyfilled && !m_isTrackingStatePolyfilled)
             {
                 return selectAction.action.WasCompletedThisFrame();
             }
@@ -208,7 +202,7 @@ namespace MixedReality.Toolkit.Input
         /// <inheritdoc />
         public float ReadValue()
         {
-            if (m_isSelectionActionValueValidCache && !GetIsPolyfillDevicePose())
+            if (!m_isSelectValuePolyfilled && !m_isTrackingStatePolyfilled)
             {
                 return selectActionValue.action.ReadValue<float>();
             }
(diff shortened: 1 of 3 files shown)


org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
```diff
@@ -145,16 +141,14 @@ protected virtual void OnDisable()
         /// </summary>
         protected virtual void Update()
         {
-            if (m_firstUpdate)
-            {
-                UpdateActionValidCaches();
-                m_firstUpdate = false;                
-            }
+            m_isTrackingStatePolyfilled = trackedPoseDriver.GetIsPolyfillDevicePose();
```
AMollis, line 144: Do we want to require a trackedPoseDriver?  In cases where the app can rely totally on action mappings, it doesn't need to pass in a traackedPoseDriver here.

I would prefer to check for null here.

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
```diff
@@ -145,16 +141,14 @@ protected virtual void OnDisable()
...
+            m_isSelectPolyfilled = !selectAction.action.HasAnyControls();
```
AMollis, line 145: Do we want to force needing a selectionAction and selectActionValue.  If the app wants to rely 100% on polyfill, these are necessarily required.   

I would prefer to check for null here...

AMollis, line 145: From what I recall...I was noticed a performance hit by calling HasAnyControls every frame.   Hence the caching being added.  

I don't think that HasAnyControls() can change often (if at all), so probably should still be caching this

AMollis (changes requested):

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
keveleigh: > Do we want to force needing a selectionAction and selectActionValue

Good question! The previous implementation did force needing these actions, but I can potentially see the use-case where an app might prefer the polyfill logic. I can null check the serialized `InputActionProperty`s here too, to account for that.

> I would prefer to check for null here...

Null-checking the action itself happens inside the `HasAnyControls()` extension method:

https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/9f1831d385880a403be4693c65331664afe5ca38/org.mixedrealitytoolkit.input/Utilities/InputActionExtensions.cs#L17-L20

keveleigh: > and I think there's an allocation in here somewhere

Only on the initial read, and we're good after that:

> After the first read, no further GC allocations should occur except if the set of controls is changed (e.g. by changing the binding mask or by adding/removing devices to/from the system).

https://docs.unity3d.com/Packages/com.unity.inputsystem@1.10/api/UnityEngine.InputSystem.InputAction.html#UnityEngine_InputSystem_InputAction_controls:~:text=After%20the%20first%20read%2C%20no%20further%20GC%20allocations%20should%20occur%20except%20if%20the%20set%20of%20controls%20is%20changed%20(e.g.%20by%20changing%20the%20binding%20mask%20or%20by%20adding/removing%20devices%20to/from%20the%20system).

> I don't think that HasAnyControls() can change often (if at all)

This is the root of the bug this PR is fixing. Controls can change as input sources are detected, connected, removed, etc. The linked doc above confirms this:

> except if the set of controls is changed (e.g. by changing the binding mask or by adding/removing devices to/from the system)

I went ahead and profiled in-editor on WMR desktop, and the `HasAnyControls()` method looks pretty cheap (0.01ms and no allocations, split across 8 calls (2 for each of the 4 in the scene) in the `PinchInputReader`):
image

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
keveleigh: For completeness, I double-checked that setting `trackedPoseDriver` to null is safe to do here. It is! The extension method correctly returns `false` here. So, I believe the proposed scenario (not needing to pass in a traackedPoseDriver) is supported.

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
```diff
@@ -145,16 +141,14 @@ protected virtual void OnDisable()
...
+            m_isSelectValuePolyfilled = !selectActionValue.action.HasAnyControls();
 
             // Workaround for missing select actions on devices without interaction profiles
             // for hands, such as Varjo and Quest. Should be removed once we have universal
```
MaxPalmer-UH, line 149: Note: Varjo uses Ultraleap hand tracking (on devices that support hand tracking, which isn't all Varjo devices). Ultraleap supports the hand interaction profile.

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
whebertML: My view is if we are now checking the validity of the select and select value actions every frame, not sure we need to also check if the tracked pose is polyfilled as well.  If either select action is not valid in a given frame, we fallback to polyfilled selection - what purpose does checking the tracked pose serve at that point?

whebertML (commented):
Changes look good to me.  I would only suggest perhaps removing the check for polyfilled tracked pose driver now that the actions are checked every frame.

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
keveleigh: Comment updated!

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
keveleigh: > If either select action is not valid in a given frame, we fallback to polyfilled selection - what purpose does checking the tracked pose serve at that point?

There exist some platforms (like Snapdragon Spaces) that don't clear a controller interaction profile even when controllers are disconnected (iirc, Quest _does_, but either path is within spec):

> The runtime **may** return the last-known interaction profile in the event that no controllers are active.

In this case, assuming the Hand Interaction Profile hasn't been added to the app's settings for whatever reason (like the way our example project is _currently_ set up, but I'll update that soon 😉), when controllers are disconnected and the user switches to hands, the select actions will still have bound controls (since the controller interaction profile is still reporting active), but the tracking state will properly relay the untracked nature of the controllers, allowing us to fall back.

My original change to support this case in XRI2 can be found at https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/704, for more info.

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
AMollis: I'm don't understand Unity's documentation with what I observe in the code.

HasAnyControls is this...
```cs
        public static bool HasAnyControls(this InputAction action)
        {
            return action?.controls.Count > 0;
        }
```

which always calls controls::get...
```cs
        public ReadOnlyArray<InputControl> controls
        {
            get
            {
                var map = GetOrCreateActionMap();
                map.ResolveBindingsIfNecessary();
                return map.GetControlsForSingleAction(this);
            }
        }
```

which always calls InputActionMap::GetControlsForSingleAction...
```cs
        internal ReadOnlyArray<InputControl> GetControlsForSingleAction(InputAction action)
        {
            Debug.Assert(m_State != null);
            Debug.Assert(m_MapIndexInState != InputActionState.kInvalidIndex);
            Debug.Assert(m_Actions != null);
            Debug.Assert(action != null);
            Debug.Assert(action.m_ActionMap == this);
            Debug.Assert(!action.isSingletonAction || m_SingletonAction == action);

            if (!controlsForEachActionInitialized)
                SetUpPerActionControlAndBindingArrays();

            return new ReadOnlyArray<InputControl>(m_ControlsForEachAction, action.m_ControlStartIndex,
                action.m_ControlCount);
        }
```

which always calls `new ReadOnlyArray`....

Is this `new` not really allocating memory????

AMollis: Ahhh ok....ReadOnlyArray is a struct.....    doh

AMollis (approved):

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
keveleigh: Ohh, good find! I didn't dig as deep into the code itself. Thank you for the analysis!

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
whebertML: @keveleigh Ah, I see.  Thanks for the explanation!  I see now where that was introduced (and why) and how this was just replacing that functionality.

PR #929, opened 2024-09-19
Add debug reference to "this" in ToggleCollection

anonymous2585:
Minor change: added debug reference to "this" in ToggleCollection.

This permit developer to click on the debug log in the console and Unity focus the component responsible for the warning.
I had this warning but couldn't know from where it was comming (I have multiples ToggleCollection in scene).


keveleigh (approved):

keveleigh:
Thanks for adding this! I think it's probably we'd benefit from adding this fairly widely throughout, so thanks for kicking off the change.

PR #933, opened 2024-09-30
Fix issue in ScaleLogic when multiple interactors have same position

### LINKED ISSUE #928: local scaling with multiple input based on far interaction
## Describe the bug

I have one game object in my scene which has an object manipulator attached. When i want to scale or move it with the controllers (Meta quest 3) everything works fine. Then when I switch to my hands, I can move the object by pinching but as soon as I want to rescale it with both hands the scene crashes and the following error appears: 
'''
transform.localScale assign attempt for '1AFB_bounding_sphere' is not valid. Input localScale is { NaN, NaN, NaN }.
UnityEngine.Transform:set_localScale (UnityEngine.Vector3)
MixedReality.Toolkit.SpatialManipulation.ObjectManipulator:ApplyTargetTransform () (at ./Library/PackageCache/org.mixedrealitytoolkit.spatialmanipulation@5fe191610f47/ObjectManipulator/ObjectManipulator.cs:882)
MixedReality.Toolkit.SpatialManipulation.ObjectManipulator:ProcessInteractable (UnityEngine.XR.Interaction.Toolkit.XRInteractionUpdateOrder/UpdatePhase) (at ./Library/PackageCache/org.mixedrealitytoolkit.spatialmanipulation@5fe191610f47/ObjectManipulator/ObjectManipulator.cs:848)
UnityEngine.XR.Interaction.Toolkit.XRInteractionManager:Update ()
'''
Here, "1AFB_bounding_sphere" is the name of my object which has the object manipulator attached. 

## To reproduce

Hardware:
- Meta quest 3 
- controllers
- your hands

Software:
- XR Hands 1.3.0
- XR Interaction Toolkit 3.0.3
- Unity OpenXR Meta 1.0.1
- Universal RP 14.0.10
- MRTK Input 4.0.0-pre.1
- MRTK Spatial Manipulation 4.0.0-pre.1 
- MRTK Standard Assets 3.2.0
- MRTK Tools 3.0.4
- MRTK Core Definitions 4.0.0-pre.1
- MK UX Components  4.0.0-pre.1
- MRTK UX Components (Non-Canvas) 4.0.0-pre.1
- MRTK Extended Assets 3.0.3 
- MRTK Diagnostics 3.0.2
- AR Foundation 5.1.5

My XR rig is basically copied from the HandsdemoScene under: Assets/Samples/XR Interaction Toolkit/3.0.3/Hands Interaction Demo/HandsDemoScene.unity. My object has the following components:
- MeshRenderer
- MeshCollider
- ObjectManipulator (with multiple selection mode and rotate about object center for near and far)

Run the scene and use both of your hands with pinching on the same object with the object manipulator to scale it.

## Expected behavior

I would have expected no error. 

- Unity Version 2022.3

- Meta quest 3 (Android)

whebertML:
Fixes #928

### DIFF
--- a/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/MoveLogics/ScaleLogic.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/MoveLogics/ScaleLogic.cs
@@ -23,8 +23,8 @@ namespace MixedReality.Toolkit.SpatialManipulation
         private Vector3 startAttachTransformScale;
         private float startHandDistanceMeters;
 
-        // Meaningful minimum distance of scaling handles to calculate scale.
-        private const float scaleDistanceEpsilon = .0001f;
+        // Meaningful minimum squared distance of scaling handles to calculate scale.
+        private const float scaleDistanceSquaredEpsilon = .0001f;
 
         /// <inheritdoc />
         public override void Setup(List<IXRSelectInteractor> interactors, IXRSelectInteractable interactable, MixedRealityTransform currentTarget)
@@ -92,7 +92,7 @@ namespace MixedReality.Toolkit.SpatialManipulation
                                                        interactors[j].GetAttachTransform(interactable).position);
                     }
                     // Ensure distance is a meaningful magnitude for scaling.
-                    distance = Mathf.Max(distance, scaleDistanceEpsilon);
+                    distance = Mathf.Max(distance, scaleDistanceSquaredEpsilon);
 
                     if (distance < result)
                     {
--- a/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/MoveLogics/ScaleLogic.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/MoveLogics/ScaleLogic.cs
@@ -78,8 +78,8 @@ namespace MixedReality.Toolkit.SpatialManipulation
                 for (int j = i + 1; j < interactors.Count; j++)
                 {
                     // Defer square root until end for performance.
-                    var distance = Vector3.SqrMagnitude(interactors[i].transform.position -
-                                                       interactors[j].transform.position);
+                    var distance = Vector3.SqrMagnitude(interactors[i].GetAttachTransform(interactable).position -
+                                                       interactors[j].GetAttachTransform(interactable).position);
                     if (distance < Mathf.Epsilon)
                     {
                         distance = 1.0f;
--- a/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/MoveLogics/ScaleLogic.cs
+++ b/org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/MoveLogics/ScaleLogic.cs
@@ -23,6 +23,9 @@ namespace MixedReality.Toolkit.SpatialManipulation
         private Vector3 startAttachTransformScale;
         private float startHandDistanceMeters;
 
+        // Meaningful minimum distance of scaling handles to calculate scale.
+        private const float scaleDistanceEpsilon = .01f;
+
         /// <inheritdoc />
         public override void Setup(List<IXRSelectInteractor> interactors, IXRSelectInteractable interactable, MixedRealityTransform currentTarget)
         {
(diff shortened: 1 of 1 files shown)


org.mixedrealitytoolkit.spatialmanipulation/ObjectManipulator/MoveLogics/ScaleLogic.cs
```diff
@@ -78,8 +78,12 @@ private float GetScaleBetweenInteractors(List<IXRSelectInteractor> interactors,
                 for (int j = i + 1; j < interactors.Count; j++)
                 {
                     // Defer square root until end for performance.
-                    var distance = Vector3.SqrMagnitude(interactors[i].transform.position -
-                                                       interactors[j].transform.position);
+                    var distance = Vector3.SqrMagnitude(interactors[i].GetAttachTransform(interactable).position -
```
keveleigh, line 81: How do we get into the case where two interactors have exactly identical positions? And is there any hint as to why it's seemingly only seen on Quest?

whebertML: @keveleigh   This only seems to occur with XRI 3 and when using the updated default XRI rig in the scene instead of our MRTK3 rig.  The issue describes reproduction using XRI's HandsDemoScene sample, and I was able to reproduce with the XRI rig instead of ours.

In the XRI 3 default rig, for the hand interactor groups, they no longer modify the position of the interactor (Near-Far Interactor) itself using something like a tracked pose driver.  Instead, that interactor utilizes a InteractionAttachController component, which creates new GameObjects to represent the attach points (anchor and child) for the interaction, themselves getting positional updates.

So, in this case, the interactors themselves not getting transform updates, had matching positions.  I know this isn't an issue with our rig, but thought it was worth protecting against cases like this where the interactor positions could match.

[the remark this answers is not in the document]
whebertML: I tested this a little and feels ok, but like you say scale is more sensitive with controller orientation.  And perhaps interactor positions were used originally to lesson that for scale?  But the rotation and position are already using the attach transforms, and rotation is definitely sensitive to that currently.

whebertML: @keveleigh @AMollis 

Also, the fact that XRI 3 is updating their systems to use things like the InteractionAttachController, which creates sub objects to represent transforms for the interaction, instead of updating the interactor transform itself, may possibly cause issues when used in combination with MRTK3's logic in other areas (like this case with ScaleLogic).

I couldn't think of other places that may be impacted off hand, but thought there might be since a moving interactor is an assumption we make generally I think.  Worth discussing?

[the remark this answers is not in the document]
whebertML: @keveleigh I tested this a bit more.  My previous test was with the XRI 3 rig and their controllers in the HandsDemoScene.  It felt mostly fine with their rig and using the controllers, but I believe they are doing some smoothing of the attach transforms.

However, when testing with the MRTK3 rig (which I should've done initially), the scaling with controller far ray is quite erratic since I don't think we smooth the far ray attach transforms much if at all.  So decided to use attach transforms only as a fallback if the interactor positions appear identical.

The only case I know of where the fallback would happen is with the XRI 3 rig's hand interactor group objects.

shaynie:
@whebertML - Please merge main back in to your PR to pick up the tools change that will let the PR build complete.

Thanks,
S

whebertML:
@shaynie Hmm, this change is on top of the feature/XRI3 branch, though, and trying to be merged to it.  I tried looking at the details of the errors, but don't have access.  What does it say?

shaynie:
@whebertML Yes, you are correct.  I'll do a PR to pull the build change to feature/XRI3.  Thanks for the catch.

shaynie:
@whebertML - here's the PR for the cherry pick to feature/XRI3 https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/935

shaynie (approved):

PR #942, opened 2024-10-15
Prevents simultaneous editing of multiple input fields when using Non-Native Keyboard

onechenxing:
Prevents simultaneous editing of multiple input fields when using Non-Native Keyboard

> Test Scene : MRTKDevTemplate/Scenes/Experimental/NonNativeKeyboard
> copying multiple MixedRealityTMPInputField causes an issue where, when multiple Input Fields are clicked, the keyboard ends up editing all of them simultaneously.
>


keveleigh (approved):
Thank you for your contribution! I tested this locally, and it looks like it works great.

PR #698, opened 2024-03-19
Adding missing [CanEditMultipleObject] attributes

ms-RistoRK:
This is for https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/issues/573

PR #943, opened 2024-10-15
Fix error message spamming when selecting PressableButtons in Editor

### LINKED ISSUE #573: [BUG] [EDITOR] Edition of multiples interactables property `IsInteractive` at the same time doesn't work
## Describe the bug

When you select multiples Pressables Buttons and check or uncheck the "IsInteractive" property, only the last one selected is actually enabled or disabled.

## To reproduce

1. Open the scene `Experimental/ScrollingExample`
2. Unfold all children of the GameObject `VerticalAndHortizontalScrolling`
3. Select all children called `Action Button (x)`
4. In the inspector, uncheck `Is Interactable`
5. See only the last selected is disabled

## Expected behavior

All selected Pressables Buttons are disabled.

## Setup

- Unity Version 2021.3.21f1
- MRTK Version or Commit: main branch [a09f5cd](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/a09f5cd98b2b47b668bcc6fc775bcc7fd3735eb7)

## Additional information

The work-around is to uncheck the Unity enabled property that is synced with the "IsInteractive" value.

I personnaly think it's better to have only one checkbox in the insperctor. The "IsInteractable" checkbox is not so useful and confusing that it also disable the component at the same time.
I'm in favor to remove it and document that a disabled button is just a disabled interactable component.

### LINKED ISSUE #772: Endless stream of error messages in Unity Editor when a GO with a PressableButton is selected
Repro:
- git checkout uxcore-v3.2.1
- Open up the MRTKDevTemplate Unity project - Unity 2022.3.20f1
- Open up the HandInteractionExamples scene
- Select a GameObject with a PressableButton component (like BasicPressableButton)

Result:
- In the Unity Console, you get a per-frame stream of "The serializedObject should not be used inside OnSceneGUI or OnPreviewGUI. Use the target property directly instead.".  This overwhelms the Unity Editor and some background tasks become very slow.

Commenting out the "[CanEditMultipleObjects]" in PressableButtonInspector.cs will make the errors go away.  Not the right fix obviously.

Looks like it was introduced in Change #698

whebertML:
* Also fixes multi-object modification of "Is Interactable" convenience alias for StatefulInteractable.enabled.

The previous fix to #573 , which was #698, supposedly caused #772.  This should fix both.

Fixes #772
Fixes #573

### DIFF
--- a/org.mixedrealitytoolkit.core/CHANGELOG.md
+++ b/org.mixedrealitytoolkit.core/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## [3.3.0-development] - 2024-06-24
+## [3.2.3-development] - 2024-06-24
 
 ### Fixed
 
-* Fixed broken project validation help link, for item 'MRTK3 profile may need to be assigned for the Standalone build target' (Issue #882) [PR#886 (https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/886)]
+* Fixed broken project validation help link, for item 'MRTK3 profile may need to be assigned for the Standalone build target' (Issue #882) [PR #886] (https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/886)
+* Fixed the "Is Interactable" convenience alias on StatefulInteractableEditor to allow multi-object editing in the Inspector to update all values. (Issue #573) [PR #943] (https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/943)
 
 ## [3.2.2-development] - 2024-06-13
 
--- a/org.mixedrealitytoolkit.core/CHANGELOG.md
+++ b/org.mixedrealitytoolkit.core/CHANGELOG.md
@@ -6,8 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
-* Fixed broken project validation help link, for item 'MRTK3 profile may need to be assigned for the Standalone build target' (Issue #882) [PR #886] (https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/886)
-* Fixed the "Is Interactable" convenience alias on StatefulInteractableEditor to allow multi-object editing in the Inspector to update all values. (Issue #573) [PR #943] (https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/943)
+* Fixed broken project validation help link, for item 'MRTK3 profile may need to be assigned for the Standalone build target' (Issue #882) [PR #886](https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/886)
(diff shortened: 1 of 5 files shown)


ms-RistoRK (dismissed):

org.mixedrealitytoolkit.core/Editor/Editors/StatefulInteractableEditor.cs
```diff
@@ -91,12 +95,9 @@ protected void DrawProperties(bool showToggleMode)
 
             StatefulInteractable interactable = target as StatefulInteractable;
 
-            bool interactableActive = EditorGUILayout.Toggle(new GUIContent("Is Interactable", "Convenience alias for StatefulInteractable.enabled"), interactable.enabled);
-
-            if (interactableActive != (target as StatefulInteractable).enabled)
+            if (IsInteractable != null)
```
anonymous2585, line 98: Curiosity: is it really necessary? In what case IsInteractable can be null?

whebertML: @anonymous2585 
You're probably right, it isn't necessary and would never be null.  However, I couldn't find any official documentation from Unity on obtaining the serialized property for .enabled, even though using "m_Enabled" is how the Editor class finds it itself.  But not being documented, that I could find, made me consider that it would be safer to add the check just in case Unity decides to change things.  The check could probably be removed, I'm cool either way.

org.mixedrealitytoolkit.uxcore/Editor/Inspectors/PressableButton/PressableButtonInspector.cs
```diff
@@ -101,7 +101,6 @@ private void OnSceneGUI()
                 return;
             }
 
-            serializedObject.Update();
```
whebertML, line 104: For future reference on removing this:

This line was originally added here:  https://github.com/microsoft/MixedRealityToolkit-Unity/pull/11296, for the issue https://github.com/microsoft/MixedRealityToolkit-Unity/issues/11295.  This issue concered the PressableButton push planes not visually updating when changed manually in the Inspector.  I tested that after removing this line and it seems fine currently.  This makes sense as `serializedObject.ApplyModifiedProperties();` is called at the end of `PressableButton.DrawProperties()`, which should apply any Inspector changes to the target object used in `OnSceneGUI()`.

org.mixedrealitytoolkit.core/CHANGELOG.md
```diff
@@ -2,11 +2,12 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## [3.3.0-development] - 2024-06-24
+## [3.2.3-development] - 2024-06-24
```
AMollis, line 5: Any clue why/how this was changed to 3.3.0 in changelong, but kept asd 3.2.2 in package.json?

AMollis: https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/commit/9f1831d385880a403be4693c65331664afe5ca38

AMollis (approved):

PR #969, opened 2025-01-09
Allow arbitrary generated folder location

### LINKED ISSUE #922: [FEATURE REQUEST] Allow moving the MRTK settings asset
## Describe the problem

MRTK3 is adding visual noise to the Unity Assets root directory by placing its MRTKSettings.asset file in the MRTK.Generated directory. Moving this directory will break building causing an error that no profile has been defined for the current platform. The error is wrong and misleading because MRTK3 expects the MRTKSettings.asset at a specific location in the Assets directory.

## Describe the solution you'd like

Previous versions of the MRTK (e.g. MRTK 2.8) had no problem moving the MRTK.Generated directory to a different place. It would be great if you could bring this back.

## Describe alternatives you've considered

Moving the directory myself but that causes a build issue.

keveleigh:
Fixes #922

Ports over the "folder sentinel" pattern from MRTK2 (https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5603)


keveleigh:
> GetOrCreateGeneratedFolderPath and AssetPostprocessor aren't run very often, right? What specifically causes this to be run? Assuming that is rare, this looks fine to me.

@whitetigerhelix AssetPostprocessor runs every time the asset database is re-imported, which shouldn't be _super_ frequent, but will happen as assets are added/removed/changed in a project. `GetOrCreateGeneratedFolderPath` also doesn't get called too frequently, but also shouldn't be doing much work, as the AssetPostprocessor should have cached the path already.

Issue #511, opened 2023-10-11
Vendor specific plugin and samples


AMollis:

shaynie:
MS MR package provides hand joint data and controller models.  Maybe not ready to remove the dependency yet.

keveleigh:
> MS MR package provides hand joint data and controller models. Maybe not ready to remove the dependency yet.

Update: hand joint data will be provided from Unity XR Hands as of #973. Controller models are still an open issue.

PR #973, opened 2025-01-10
Update MRTK Default Profile to use Unity's XR Hands package

keveleigh:
Part of #66
Part of #511


MaxPalmer-UH:
Seems OK testing with the HandInteractionExamples scene on a Quest 3 with Meta's hand tracking. Note, not an MRTK issue but interesting I see this warning from Unity, which doesn't seem appropriae.
image

MaxPalmer-UH (dismissed):

keveleigh:
@MaxPalmer-UH I went ahead and added some MRTK3 validation rules to hopefully help with that confusing validation state:

{710FF962-0625-4D40-9507-1D0409F26C44}

MaxPalmer-UH (approved):
Thanks for that addition to check/fix if the Subsystem for OpenXR hands is enabled. Seems to work fine.

PR #974, opened 2025-01-15
Update CHANGELOG.md for UX Core

keveleigh:
Whoops, forgot to update the changelog in #942

### DIFF
--- a/org.mixedrealitytoolkit.uxcore/CHANGELOG.md
+++ b/org.mixedrealitytoolkit.uxcore/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## [3.3.0-development] - 2024-09-12
+## Unreleased
 
 ### Changed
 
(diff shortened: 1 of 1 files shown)


org.mixedrealitytoolkit.uxcore/CHANGELOG.md
```diff
@@ -11,21 +11,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
-## [3.2.2-development] - 2024-08-29
+## [3.2.2] - 2024-09-18
 
 ### Changed
 
-* Package patch version update to allow UPM publishing
+* Package patch version update to allow UPM publishing.
 
-## [3.2.1] - 2024-04-23
+## [3.2.1] - 2024-06-12
```
whebertML, line 22: These dates should be the tag/release dates for 3.2.2 and 3.2.1?  Perhaps missing something, but those aren't 2024-09-11 and 2024-06-03?

keveleigh: I was going based on the dates on the [Unity Asset Store](https://assetstore.unity.com/packages/add-ons/mrtk-ux-core-scripts-286979#releases) (where relevant) or the [NPM feed](https://dev.azure.com/aipmr/MixedReality-Unity-Packages/_artifacts/feed/Unity-packages/Npm/org.mixedrealitytoolkit.uxcore/versions/4.0.0-pre.1) (if the version isn't on the Asset Store).

But, we should probably establish some agreed-upon convention, seeing as you've pointed out an additional source of dates! I don't have a strong preference here, but was defaulting to the date the packages were published to the feeds.

whebertML: Ah, right, totally forgot to look at those dates.

I'm cool with whatever convention, but seeing how Unity will be/should be? the preferred method eventually, totally fine with that being it.  Something good to bring up at next meeting.  Approving.

whebertML (approved):

PR #976, opened 2025-01-29
Introduce ScaleMaintainType

### LINKED ISSUE #979: [FEATURE REQUEST] Need a more advanced way of scaling the bounding box handles
## Describe the problem

Currently the bounding box handles scale based on the initial parent scale. We have scenarios where items can be scaled and then saved with the new scale or your bounding box can be transferred from one object to another. Since the scale of the handles is initialized on awake this can lead to different handle scales at different times, XL or XS handles when swapping, and an overall not deterministic experience. Additionally, we would like min and max lossy scale values that the handles will stay at when they reach those points.

## Describe the solution you'd like

We've developed a solution where you can set the type of scaling you want and created a new type that is deterministic how we would like no matter what the initial parent scale is. Instead of just having to go with the scaling that has been written the user can select which type of scaling they would like or add their own type. We also added the Advanced type that we will be using to address these issues.

## Describe alternatives you've considered

We created a script to overwrite the scale every frame but this is not performant

## Additional context

Image

nayanatharap:
Fixes #979 - Introduce different options for scaling bounds handles. The options include:
- GlobalSize - Maintains global size (previously stored as a separate bool property):
GlobalSize

- FixedScale - Sets scale based on initial parent scale:
FixedScale

- Advanced - Adjusts scale to retain same size
Advanced


nayanatharap:
1. GlobalSize: 
GlobalSize

2. FixedScale
FixedScale

3. Advanced
Advanced

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
```diff
@@ -36,13 +36,34 @@ public BoundsControl BoundsControlRoot
         }
 
         [SerializeField]
-        [Tooltip("Should the handle maintain its global size, even as the object changes size?")]
-        private bool maintainGlobalSize = true;
+        [Tooltip("How should the handle scale be maintained?")]
+        private ScaleMaintainType scaleMaintainType = ScaleMaintainType.GlobalSize;
 
+        // For backward compatibilty, the scaleAdjustType is set according to the maintainGlobalSize property on start.
```
keveleigh, line 42:
```suggestion
        // For backward compatibility, the scaleAdjustType is set according to the maintainGlobalSize property on start.
```

org.mixedrealitytoolkit.spatialmanipulation/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "org.mixedrealitytoolkit.spatialmanipulation",
-  "version": "3.3.1-development",
+  "version": "3.3.2-development",
```
keveleigh, line 3: Since we're adding a new feature here, we probably want to jump to 3.4.0

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
```diff
@@ -36,13 +37,53 @@ public BoundsControl BoundsControlRoot
         }
 
         [SerializeField]
-        [Tooltip("Should the handle maintain its global size, even as the object changes size?")]
+        [Tooltip("How should the handle scale be maintained?")]
+        private ScaleMaintainType scaleMaintainType = ScaleMaintainType.GlobalSize;
+
+        // A temporary variable used to migrate instances of FollowJoint to use the jointPoseSource class as the source of truth
```
keveleigh, line 43: Looks like this comment needs updating to refer to this script!

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
```diff
@@ -131,27 +180,67 @@ protected virtual void LateUpdate()
...
+                    // the handles grow/shrink with the overall box manipulation.
+                    transform.localScale = targetScale * (transform.parent.lossyScale.MaxComponent() / targetParentScale);
+
+                    // If this scale is greater than our desired lossy scale then clamp it to the max lossy scale
+                    if (transform.lossyScale.MaxComponent() > maxLossyScale)
+                    {
+                        transform.localScale = maxScale;
+                    }
+                    // If this scale is less than our desired lossy scale then clamp it to the min lossy scale
+                    else if (transform.lossyScale.MinComponent() < minLossyScale)
```
keveleigh, line 226: This may just be misunderstanding the flow here, but do we want this to be an `else if` or should it just be a second `if` statement? Is the min component clamping affected by the max component clamping, or are they independent?

keveleigh (approved):
Approved, pending my last question!

org.mixedrealitytoolkit.spatialmanipulation/BoundsControl/BoundsHandleInteractable.cs
nayanatharap: Since we are setting x, y, z to all be the same exact lossy scale, having both >max and <min wouldn't be possible (unless the component was 0, in which case it would only be the first frame). So `else if` vs `if` doesn't make a difference in this case.

whitetigerhelix (approved):

PR #977, opened 2025-01-30
Update .gitignore to include .utmp folder

keveleigh:
With Unity 6, I've started seeing this folder generated when creating Android builds. Adding it to the .gitignore.


whebertML (approved):
Saw that as well with Unity 6.

PR #984, opened 2025-02-19
Update controller model visualization to hide when hands become tracked after initialization

keveleigh:


ms-RistoRK (commented):
LGTM, unfortunately I no longer have access to a HL2 device to test.

whitetigerhelix (approved):

PR #978, opened 2025-01-30
Don't register an unsupported Windows speech subsystem

keveleigh:
`WindowsKeywordRecognitionSubsystem` now checks itself before attempting registration to ensure it'll run on a supported platform.

Fixes this exception on non-Windows platforms:

{1EA4AD3A-10E9-4DF9-A502-6850FF16450E}

PR #1001, opened 2025-03-28
Don't register dictation subsystem on non-Windows

keveleigh:
Follow-up to https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/978


whitetigerhelix (approved):

RCrolyFex:
I think a tarball release would be great, I haven't switched over to the asset store method yet

keveleigh:
@RCrolyFex Released! https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/releases/tag/core-v3.3.0

Please let me know if you run into any issues! It's been a while since the last release, so this one was on the larger side 😅 hoping nothing fell through the cracks!

Issue #1006, opened 2025-04-15
[BUG] SliderSounds play sounds even when disabled


anonymous2585:
## Describe the bug

We want to drive a slider from a script, updating its value without playing sound. For that, we disable the SliderSound script before changing the value and re-enabling it after.
It doesn't work because SliderSounds always play sounds even after being disabled.

Edit: I saw the "Play Sounds Only On Interact" option only after making this issue. It work great for my usecase, but the script still shouldn't play sound when disabled.

## To reproduce

1. Open scene "HandInteractionExamples"
2. Press play
3. Move the slider above the "Pinch Interaction" title, see it plays sound (it's ok)
4. Go in hierarchy, disable the script "SliderSounds" on GameObject "MixedRealitySceneContent/Touch Interaction/Slider_168x40mm_IconAndValue/VisualRoot/ScaleRoot/Content/NonCanvasSliderBase".
5. Move the slider above "Pinch Interaction" title, see it still plays sound (it's ko)

## Expected behavior

Slider doesn't play sound when the "SliderSounds" is disabled.

## Your setup

- Unity Version 6000.0.37f1
- MRTK UX Core 4.0.0-development.pre.1

keveleigh:
Fixed with #1007. Thank you!!

PR #1004, opened 2025-04-10
Add a build flag for building with hand joint debugging hands

keveleigh:
This allows CI to generate APKs for validating the hand joint locations.
This PR also adds menu items for patching in-editor:
{1D94A6EA-DD89-4863-8A9D-4C3EA6F37851}

This is largely an example project QoL change!

Debug:
{9F581E82-C5EC-4541-9D92-45C99886BCAA}

Non-debug:
{830B1D62-ED11-456C-B8D8-C48F83046F65}

PR #1007, opened 2025-04-15
Fix SliderSounds playing sound even when disabled (#1006)

anonymous2585:
- Unsubscribe SliderSounds from Slider event when disabled
- Updated utility function EnsureComponent to use TryGetComponent

### DIFF
--- a/org.mixedrealitytoolkit.core/Utilities/Extensions/ComponentExtensions.cs
+++ b/org.mixedrealitytoolkit.core/Utilities/Extensions/ComponentExtensions.cs
@@ -48,8 +48,7 @@ namespace MixedReality.Toolkit
         /// </remarks>
         public static T EnsureComponent<T>(this GameObject gameObject) where T : Component
         {
-            T foundComponent = gameObject.GetComponent<T>();
-            return foundComponent == null ? gameObject.AddComponent<T>() : foundComponent;
+            return gameObject.TryGetComponent(out T foundComponent) ? foundComponent : gameObject.AddComponent<T>();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


org.mixedrealitytoolkit.core/Utilities/Extensions/ComponentExtensions.cs
```diff
@@ -48,8 +48,7 @@ public static T FindAncestorComponent<T>(this Component component, bool includeS
         /// </remarks>
         public static T EnsureComponent<T>(this GameObject gameObject) where T : Component
         {
-            T foundComponent = gameObject.GetComponent<T>();
-            return foundComponent == null ? gameObject.AddComponent<T>() : foundComponent;
+            return gameObject.TryGetComponent(out T foundComponent) ? foundComponent : gameObject.AddComponent<T>();
```
keveleigh, line 51: Thank you for this! I made the same change in https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/pull/1004/files#diff-9827b8e3e0230bbf3ccacee31f3b8a03d51c41e428ff954d60a08699353580f1R51, which is still pending review, so I'm happy to go ahead and merge yours!

org.mixedrealitytoolkit.uxcore/Slider/SliderSounds.cs
```diff
@@ -127,28 +127,39 @@ public class SliderSounds : MonoBehaviour
...
+        private void OnEnable()
+        {
             slider.firstSelectEntered.AddListener(OnInteractionStarted);
             slider.lastSelectExited.AddListener(OnInteractionEnded);
             slider.OnValueUpdated.AddListener(OnValueUpdated);
         }
 
+        private void OnDisable()
+        {
+            slider.firstSelectEntered.RemoveListener(OnInteractionStarted);
```
keveleigh, line 158: Oof, thanks for catching this!

keveleigh (approved):

Issue #1005, opened 2025-04-11
[FEATURE REQUEST] Input Simulator: camera movement speed famerate dependency fix


UniversalLanguageAlgorithm:
## Describe the problem
We're using the MRTK Input Simulator to demo our XR app on the desktop and WebGL platforms. 
Currently, SimulatedHMD.cs doesn't scale the resulting camera movement speed with Time.deltaTime if "Input" is not smoothed. Or, if "Input" is smoothed, it just uses fixedDeltaTime. 

This makes the resulting maximum speed depend on the framerate causing drastically different movement speeds not only between Editor and Player, but also between different various resolutions in Play Mode. On top of that, all of these difference depend on the hardware capabilities executing the app. 

See line below:
[https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity/blob/4f1ebac12e59d2110ff0bd9a5e9e9281af394eee/org.mixedrealitytoolkit.input/Simulation/Devices/SimulatedHMD.cs#L143](url)  

We are aware that the situation is the result of an editor-only bug in Unity that makes Time.deltaTime unreliable when used with certain input devices.

## Describe the solution you'd like
The comment on line 130 states that the current behavior is intentional. However, the way we understand this is that the worst thing to happen is this: input-device induced lagspikes which render Time.deltaTime unreliable will only defeat the smoothing of `MoveDelta`. Since that is used essentially as a speed variable, it means smoothed starts or smoothed stops will happen instantly.
We believe that having a consistent max speed across editor and builds should take priority over smoothed acceleration and would suggest the following:  

1. Change line 136
 `Smoothing.SmoothTo(smoothedMoveDelta, moveDelta, moveSmoothingTime, Time.fixedDeltaTime) :`
to
`Smoothing.SmoothTo(smoothedMoveDelta, moveDelta, moveSmoothingTime, Time.deltaTime) :`

2. Change line 143: 
` Vector3 cameraPosition = simulatedHmdState.centerEyePosition + Quaternion.Inverse(origin.rotation) * Camera.main.transform.rotation * (smoothedMoveDelta * moveSpeed);`
to 
` Vector3 cameraPosition = simulatedHmdState.centerEyePosition + Quaternion.Inverse(origin.rotation) * Camera.main.transform.rotation * (smoothedMoveDelta  * moveSpeed*Time.deltaTime * 1000f);`
[...]

PR #1011, opened 2025-04-24
Add option for framerate independent movement in Input Simulator

UniversalLanguageAlgorithm:
Implement changes suggested in FR #1005 to ensure consistent movement speed when using simulated input. 
Add option to make SimulatedHMD multiply camera movement speed with deltaTime, which has been originally ommited due to the Editor's mouse polling bug invalidating Time.deltaTime. 
Added a "Framerate Independent" checkbox in the Input Simulator inspector, to disable it if needed.


keveleigh:
Thanks for this contribution! Looks generally good to me. I just pushed up a few docs changes and other minor updates, but I'm happy to approve. Thanks again!

PR #1041, opened 2025-07-11
Improve fallback logic in PinchInputReader.cs

keveleigh:
* Changed `hasPinchData` to `isInProgress`, to better represent the Unity method `IsInProgress()` when going through the fallback path
* Updated the polyfill logic to check if one of the "select" and "select value" is bound while the other is polyfilled. We can use the data from the bound action to synthesize the other better than the hand joint logic will. This also helps account for controller interaction profiles, like `khr/simple_controller`, that doesn't have both a bool and float action for select.

### DIFF
--- a/org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
+++ b/org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
@@ -29,7 +29,7 @@ namespace MixedReality.Toolkit.Input
         /// </summary>
         private struct FallbackState
         {
-            public bool hasPinchData;
+            public bool isInProgress;
             public bool isPerformed;
             public bool wasPerformedThisFrame;
             public bool wasCompletedThisFrame;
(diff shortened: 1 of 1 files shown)


org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
```diff
@@ -221,11 +212,9 @@ public bool TryReadValue(out float value)
...
                 return action.IsInProgress();
             }
-            else
-            {
-                value = m_fallbackState.value;
-                return m_fallbackState.hasPinchData;
-            }
+
+            value = m_fallbackState.value;
+            return value > 0;
```
keveleigh, line 217: FYI @whebertML, wanted to make sure this doesn't affect the intent of #852

org.mixedrealitytoolkit.input/Readers/PinchInputReader.cs
```diff
@@ -242,6 +231,23 @@ private void UpdatePinchSelection()
...
+                // This section accounts for one of "select" and "select value" being bound while the other is polyfilled.
+                // We can use the data from the bound action to synthesize the other better than the hand joint logic will.
+                if (!m_isSelectPolyfilled && !m_isTrackingStatePolyfilled)
+                {
+                    m_fallbackState.value = ReadIsPerformed() ? 1 : 0;
+                    return;
+                }
+                else if (!m_isSelectValuePolyfilled && !m_isTrackingStatePolyfilled)
+                {
+                    bool isPinched = ReadValue() >= (m_fallbackState.isPerformed ? 0.9f : 1.0f);
```
whitetigerhelix, line 243: nit, maybe we make these constants instead of hardcoding?

whitetigerhelix (approved):

PR #1080, opened 2025-11-19
Added some missing changelog entries and moved some into their correct releases

keveleigh:
Added some missing changelog entries and moved some into their correct releases.


simon-ign (approved):
Verified all changelog additions are merged and link correctly

PR #1119, opened 2026-05-20
Theming!

keveleigh:
# MRTK UX Theming

## Overview

MRTK Theming 3.0 is a lightweight data-binding system that routes named values from a theme asset to components in the scene. Switching the active theme on a `ThemeDataSource` instantly updates every bound component across all scenes and prefabs that reference it.

The system has three layers:

| Layer | Asset / Component | Purpose |
|---|---|---|
| **Schema** | `ThemeDefinition` (ScriptableObject) | Declares what named items exist and what type each one is |
| **State** | `ThemeDataSource` (ScriptableObject) | The single source of truth for the currently active theme |
| **Values** | `Theme` (ScriptableObject) | Provides a concrete value for every item in the definition |
| **Binding** | `ThemeBinding` (MonoBehaviour) + binders | Connects a theme item to a specific component on a GameObject |

> [!NOTE]
> `ThemeBinding` exists in the UX Core package for cross-package compatibility.

---

## Theme data source

A `ThemeDataSource` ScriptableObject is the single source of truth for which theme is currently active. Create one per logical theme domain (e.g. one for the whole app, or one per scene).

The `ThemeDataSource` holds:

- **`activeTheme`** — the currently applied `Theme` asset.
- **`themeDefinition`** — the `ThemeDefinition` asset (schema) that all compatible themes must conform to.

To switch themes at runtime, assign a new `Theme` to `themeDataSource.value`. Every `ThemeBinding` component subscribed to that source will immediately receive and apply the new values.

### Theme definition

A `ThemeDefinition` ScriptableObject is a list of named items, each with a declared data type. It acts as the contract between the `ThemeDataSource` and all `Theme` assets that reference it — every `Theme` must provide a value for every item in the definition.

Each item has:

- **Name** — a string key used to match binders to values at runtime (e.g. `"Primary Color"`, `"Backplate Material"`).
- **Data type** — a concrete subclass of `BaseThemeItemData<T>` (e.g. `ColorThemeItemData`, `MaterialThemeItemData`).

---

## Theme

A `Theme` ScriptableObject is assigned to a `ThemeDataSource` and provides a concrete value for every item declared in its referenced `ThemeDefinition` schema.

The `Theme` inspector automatically displays one field per definition item, typed correctly. Themes are **complete** — every item in the definition must have a value. This ensures that switching themes always fully overwrites every bound property, with no leftover values from a previous theme.

---

## Theme binding

The `ThemeBinding` MonoBehaviour connects a `ThemeDataSource` to one or more components on the same GameObject (or its children). Add it to any GameObject that should respond to theme changes.

`ThemeBinding` holds:

- **`themeDataSource`** — the source to subscribe to.
- **`binders`** — a list of `IBinder` implementations, each connecting one theme item to one component.

On `OnEnable`, each binder subscribes to the data source and immediately receives the current theme. On `OnDisable`, binders unsubscribe.

### Theme binders

A theme binder connects a named theme item to a specific component property. Each binder has:

- **Target** — the component to update (e.g. a `Graphic`, `Transform`, `Renderer`).
- **Bound Theme Item** — the name of the item in the `ThemeDefinition` to read from.

When the active theme changes, the binder looks up its named item in the new theme and calls `Apply` with the value.

#### Built-in binders

| Binder | Data type | Target | Sets |
|---|---|---|---|
| `GraphicColorBinder` | `Color` | `Graphic` | `Graphic.color` |
| `GraphicMaterialBinder` | `Material` | `Graphic` | `Graphic.material` |
| `RendererMaterialBinder` | `Material` | `Renderer` | `Renderer.material` |
| `TMP_TextFontAssetBinder` | `TMP_FontAsset` | `TMP_Text` | Font asset |
| `ImageSpriteBinder` | `Sprite` | `Image` | `Image.sprite` |
| `SpriteRendererSpriteBinder` | `Sprite` | `SpriteRenderer` | `SpriteRenderer.sprite` |
| `SpriteRendererColorBinder` | `Color` | `SpriteRenderer` | `SpriteRenderer.color` |
| `TransformLocalScaleBinder` | `Vector3` | `Transform` | `Transform.localScale` |
| `LayoutGroupPaddingBinder` | `RectOffset` | `HorizontalOrVerticalLayoutGroup` | Padding |
| `RoundedRectMask2DRadiusBinder` | `float` | `RoundedRectMask2D` | Corner radius |
| `FontIconSetBinder` | `FontIconSetData` | — | Font + icon set |
| `StateVisualizerEffectColorBinder` | `Color` | `StateVisualizer` | Tint effect color per state |

---

### Creating a new theme item data type

Subclass `BaseThemeItemData<T>` with the desired value type and mark it `[Serializable]`:

```csharp
[Serializable]
public class FloatItemData : BaseThemeItemData<float> { }
```

The new type will automatically appear in the `ThemeDefinition` item type dropdown in the Inspector.

### Creating a new binder

Subclass `BaseThemeBinder<T, K>` where `T` is the value type and `K` is the target component type. Mark it `[Serializable]` and implement `Apply`:

```csharp
[Serializable]
public class MyCustomBinder : BaseThemeBinder<Color, MyComponent>
{
    protected override void Apply(BaseThemeItemData<Color> themeItemData)
    {
        if (Target != null)
        {
            Target.MyColorProperty = themeItemData.Value;
        }
    }
}
```

The new binder will automatically appear in the `ThemeBinding` binder type dropdown in the Inspector, and the **Bound Theme Item** dropdown will show only the definition items whose data type is compatible with `T`.

---

## How theme switching works at runtime

1. `themeDataSource.value = newTheme` is called.
2. `ThemeDataSource` fires a `ChangeEvent<Theme>` to all listeners.
3. Each `ThemeBinding` component's binders receive the event via `OnThemeChanged`.
4. Each binder calls `Theme.TryGetItemData(ThemeDefinitionItemName, ...)` to look up its value.
5. If found, `Apply` is called with the new value. If not found, a warning is logged.

Because themes are complete (every definition item has a value in every theme), every bound property is always overwritten on a theme switch — there are no partial updates or leftover values from a previous theme.


simon-ign (approved):

org.mixedrealitytoolkit.theming/ItemData/RectOffsetItemData.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Mixed Reality Toolkit Contributors
+// Licensed under the BSD 3-Clause
+
+using UnityEngine;
+
+namespace MixedReality.Toolkit.Theming
+{
+    public class RectOffsetItemData : BaseThemeItemData<RectOffset> { }
```
joschemd-MS, line 9: With Unity 6 behaviour changes, and this class potentially being used in SerializedReference, we should add the  System.Serializable attribute to it

org.mixedrealitytoolkit.uxcore/Theming/ThemeBinding.cs
```diff
@@ -0,0 +1,57 @@
...
+        protected void OnEnable()
+        {
+            if (themeDataSource == null)
+            {
+                Debug.LogWarning($"{nameof(ThemeBinding)} on '{gameObject.name}' has no {nameof(ThemeDataSource)} assigned.", this);
+                return;
+            }
+
+            foreach (IBinder binder in binders)
+            {
```
joschemd-MS, line 37: This will throw a null reference exception if binders is not initialized above. Add a check.

org.mixedrealitytoolkit.uxcore/Theming/ThemeBinding.cs
```diff
@@ -0,0 +1,57 @@
...
+                    Debug.LogWarning($"{nameof(ThemeBinding)} on '{gameObject.name}' has a {binder.GetType().Name} with no theme item assigned.", this);
+                }
+
+                binder.Subscribe(themeDataSource);
+            }
+        }
+
+        protected void OnDisable()
+        {
+            foreach (IBinder binder in binders)
```
joschemd-MS: Same here, add a null check

org.mixedrealitytoolkit.theming/Theming/Theme.cs
```diff
@@ -0,0 +1,76 @@
...
+        private ThemeDataSource definition;
+
+        /// <summary>
+        /// The schema that this theme conforms to. Every theme item must map to an item defined here.
+        /// </summary>
+        public ThemeDataSource Definition => definition;
+
+        [SerializeReference]
+        [Tooltip("The items defining this theme's data mapped to the definition's items.")]
+        private List<ThemeItem> themeItems;
```
joschemd-MS: if this is iterated on, and not initialized, will cause issues. Recomendation: it with an empty instance or ensure there's a null guard

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
```diff
@@ -0,0 +1,131 @@
...
+                            else
+                            {
+                                // Fallback for any item whose Data doesn't follow the
+                                // BaseThemeItemData<T> shape (e.g. a null reference).
+                                EditorGUILayout.PropertyField(themeItemProp, true);
+                            }
+                        }
+                    }
+                }
+                themeItemsProp.arraySize = themeDefinitionArrayProp.arraySize;
```
joschemd-MS: This should be investigated, 1) we can get into a situation that the themItemsProp (if not properly intialized can be null, see comment in Theme.cs) and should we modifying it if we are processing the foldout (see line 100)?

Do we need to resize it if we are not processing the foldout?

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
```diff
@@ -0,0 +1,131 @@
...
+                                    themeItemsProp.MoveArrayElement(existingIndex, i);
+                                }
+                                else
+                                {
+                                    // No existing item found — insert a fresh one with default values.
+                                    string valueDataType = themeDefinitionItem.FindPropertyRelative(InspectorUIUtility.GetBackingField(nameof(ThemeDefinition.ThemeDefinitionItem.DataType))).FindPropertyRelative("reference").stringValue;
+
+                                    themeItemsProp.InsertArrayElementAtIndex(i);
+                                    themeItem = themeItemsProp.GetArrayElementAtIndex(i);
+                                    themeItem.managedReferenceValue = new Theme.ThemeItem(themeDefinitionItemName, Activator.CreateInstance(Type.GetType(valueDataType)));
```
joschemd-MS: Should have a null check on the Type.GetType(valueDataType) and handled appropriately before using it in Activator.CreateInstance.

org.mixedrealitytoolkit.theming/Editor/BaseThemeBinderDrawer.cs
```diff
@@ -0,0 +1,143 @@
...
+                {
+                    EditorGUI.LabelField(labelRect, "Bound Theme Item");
+                    selected = EditorGUI.Popup(controlRect, selected, System.Array.ConvertAll(names.ToArray(), n => new GUIContent(n)));
+                    if (check.changed)
+                    {
+                        themeDefinitionItemName.stringValue = names[selected];
+                    }
+                }
+            }
+            else if (themeDefinitionItemName != null && names == null)
```
joschemd-MS: This branch fires whenever themeDataSource == null or the data source has no matching items for that binder type. If the user removes (or temporarily clears) the ThemeDataSource reference on the ThemeBinding, every binder in the array immediately loses its saved ThemeDefinitionItemName, and the change is committed by the ApplyModifiedProperties(). Recommendation: Remove this branch, even if the datasource is unavailable, clearing the name has no benefit

org.mixedrealitytoolkit.theming/Theming/ThemeDataSource.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        #region INotifyValueChanged<Theme>
+
+        /// <inheritdoc/>
+        public Theme value
+        {
+            get => activeTheme;
+            set
+            {
+                if (value != null && value.Definition != this)
```
joschemd-MS: what happens with a brand new unconfigured theme asset? the Definition technically be null and valid correct?

org.mixedrealitytoolkit.theming/Theming/ThemeDataSource.cs
```diff
@@ -0,0 +1,69 @@
...
+                {
+                    Debug.LogError($"New theme's definition ({value.Definition.name}) does not match this data source's active definition ({name})");
```
joschemd-MS: If Definition is null, this will throw a NullReference Exception

org.mixedrealitytoolkit.theming/Editor/ThemeDataSourceEditor.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            while (iterator.NextVisible(enterChildren))
+            {
+                enterChildren = false;
+
+                EditorGUILayout.PropertyField(iterator, true);
+
+                if (iterator.name == "activeTheme" && iterator.boxedValue is Theme theme && theme.Definition != serializedObject.targetObject)
+                {
+                    EditorGUILayout.HelpBox($"Assigned theme's definition ({theme.Definition.name}) does not match this data source's active definition ({serializedObject.targetObject.name}).\nThis will lead to undefined behavior at runtime.", MessageType.Error);
```
joschemd-MS: Missing Null check on Definition.Name

org.mixedrealitytoolkit.theming/Binders/FontIconSetBinder.cs
```diff
@@ -0,0 +1,45 @@
...
+using UnityEngine;
+
+namespace MixedReality.Toolkit.Theming
+{
+    [Serializable]
+    public class FontIconSetBinder : BaseThemeBinder<FontIconSetBinder.FontIconSetData, FontIconSetBinder.FontIconSetBinding>
+    {
+        protected override void Apply(BaseThemeItemData<FontIconSetData> themeItemData)
+        {
+            if (Target != null && Target.TextMeshProComponent != null && Target.IconSelector != null && themeItemData.Value.FontIconSet.GlyphIconsByName.ContainsKey(Target.IconSelector.CurrentIconName))
```
joschemd-MS: ThemeItemData maybe null here, so can the FontIconSet , and the reference or property Target.IconSelector.CurrentIconName .. Add null check guards and handle appropriately

org.mixedrealitytoolkit.uxcore/Theming/ThemeBinding.cs
```diff
@@ -0,0 +1,57 @@
...
+                    Debug.LogWarning($"{nameof(ThemeBinding)} on '{gameObject.name}' has a null binder entry.", this);
+                    continue;
+                }
+
+                if (string.IsNullOrWhiteSpace(binder.ThemeDefinitionItemName))
+                {
+                    Debug.LogWarning($"{nameof(ThemeBinding)} on '{gameObject.name}' has a {binder.GetType().Name} with no theme item assigned.", this);
+                }
+
+                binder.Subscribe(themeDataSource);
```
joschemd-MS: This worries me, but may not be an issue, This smells of a leak, if anything else modifies the current item, but an editor or something else update binders (editor, reflection, domain reload etc) this may not properly be unsubscribed.
Recomendation: keep a cache of all items that are subscribed in the OnEnable and also properly handled them in OnDisable

org.mixedrealitytoolkit.uxcore/StateVisualizer/StateVisualizer.cs
```diff
@@ -352,9 +352,85 @@ private void OnDestroy()
         }
 
         /// <summary>
-        /// Adds the provided effect to the state with name <paramref name="stateName"/>.
-        /// Creates the state if it doesn't exist.
+        /// Sets the tint color on all <see cref="ITintEffect"/> effects within the named state.
```
joschemd-MS, line 355: Looks like &lt;/summary&gt; was removed

joschemd-MS (commented):
Please review

org.mixedrealitytoolkit.theming/Editor/BaseThemeBinderDrawer.cs
keveleigh: Oof, good catch! Removed

org.mixedrealitytoolkit.theming/Theming/Theme.cs
keveleigh: Added explicit initialization and null guards.

org.mixedrealitytoolkit.uxcore/Theming/ThemeBinding.cs
keveleigh: Added explicit initialization and null guards.

org.mixedrealitytoolkit.uxcore/Theming/ThemeBinding.cs
keveleigh: Added explicit initialization and null guards.

org.mixedrealitytoolkit.theming/ItemData/RectOffsetItemData.cs
keveleigh: Added!

org.mixedrealitytoolkit.theming/Theming/ThemeDataSource.cs
keveleigh: Updated this to print "null" if Definition is null

org.mixedrealitytoolkit.theming/Editor/ThemeDataSourceEditor.cs
keveleigh: Updated this to print "null" if Definition is null

org.mixedrealitytoolkit.uxcore/StateVisualizer/StateVisualizer.cs
keveleigh: Oops re-added

org.mixedrealitytoolkit.theming/Binders/FontIconSetBinder.cs
keveleigh: Added all the null checks I could think of!

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
keveleigh: Good callouts! Yeah, I think I was being a bit overzealous by resizing in all cases. It missed some initialization logic that only happens when the foldout is open.

I refactored this whole logic out to be more intentional about when it's called, what it touches, and whatnot.

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
keveleigh: Added!

org.mixedrealitytoolkit.theming/Theming/ThemeDataSource.cs
keveleigh: Good call! Updated this to check against null.

org.mixedrealitytoolkit.theming/Theming/ThemeDataSource.cs
keveleigh: Actually, this is no longer an issue after the new null check added above

org.mixedrealitytoolkit.uxcore/Theming/ThemeBinding.cs
keveleigh: Oof, good catch! Definitely a possibility. Added some caches here

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
```diff
@@ -0,0 +1,152 @@
...
+                    {
+                        // No existing item found — insert a fresh one with default values.
+                        string valueDataType = themeDefinitionItem.FindPropertyRelative(DataTypeField).FindPropertyRelative("reference").stringValue;
+
+                        themeItemsProp.InsertArrayElementAtIndex(i);
+                        themeItem = themeItemsProp.GetArrayElementAtIndex(i);
+
+                        Type dataType = Type.GetType(valueDataType);
+                        object dataInstance = dataType != null ? Activator.CreateInstance(dataType) : null;
+                        themeItem.managedReferenceValue = new Theme.ThemeItem(themeDefinitionItemName, dataInstance);
```
joschemd-MS: Still can possibly create a ThemeItem with a null dataInstance here. You'll end up with log spam if not corrected. What's the proper fall back expected?

org.mixedrealitytoolkit.theming/Editor/ThemeDataSourceEditor.cs
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    SerializedProperty definitionProp = serializedObject.FindProperty("themeDefinition");
+                    UnityEngine.Object activeDefinition = definitionProp?.objectReferenceValue;
+
+                    if (theme.Definition != activeDefinition)
+                    {
+                        EditorGUILayout.HelpBox($"Assigned theme's definition ({(theme.Definition != null ? theme.Definition.name : "null")}) does not match this data source's " +
+                            $"active definition ({(activeDefinition != null ? activeDefinition.name : "null")}).\nThis will lead to undefined behavior at runtime.", MessageType.Error);
+                    }
+                }
```
joschemd-MS: This caught my eye, a couple of questions:
Theme and ThemeDataSource both reference ThemeDefinition correct?
If so ThemeDefinition could be a scriptable object so this comparison is more accurate and the same reference will make line 32 be more accurate and allow the schema to actually be shared (instead of new save instances of the same data)

otherwise recommend code change:

if (iterator.name == "activeTheme" && iterator.objectReferenceValue is Theme theme
	&& theme.Definition != null && theme.Definition != serializedObject.targetObject)
{
	EditorGUILayout.HelpBox($"Assigned theme's definition ({theme.Definition.name}) does not match this data source ({serializedObject.targetObject.name}).\nThis will lead to undefined behavior at runtime.", MessageType.Error);
}

keveleigh: > If so ThemeDefinition could be a scriptable object so this comparison is more accurate and the same reference will make line 32 be more accurate and allow the schema to actually be shared (instead of new save instances of the same data)

Ah yeah, great callout. Originally, there _were_ three Scriptable Objects: **Theme**, **Theme Definition**, and **Theme Data Source**.

I squashed Theme Definition and Theme Data Source together, as it helped minimize the permutations I had to worry about.

You do make a good point about sharing the schema though. I'll see how easy it is to split back out.

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
keveleigh: I think `null` here is a valid, albeit ideally temporary, outcome. The fix will need to be in the Theme Definition, changing to type to one that's valid. I've updated the logic here to cache any failed types and continue checking if a new, non-null type becomes available for serialization. This is how this case looks across all relevant editors:

Theme Definition:
image

Theme:
image

I also audited the case on ThemeBinding where a binder is selected with no corresponding types in the Definition/Theme:
image

org.mixedrealitytoolkit.theming/Editor/ThemeDataSourceEditor.cs
keveleigh: Split this back out! Feels architecturally better, so thanks for pushing back on that.

org.mixedrealitytoolkit.uxcore/Theming/MRTK.UXCore.Theming.asmdef
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "MixedReality.Toolkit.UXCore.Theming",
+    "rootNamespace": "MixedReality.Toolkit.UX",
+    "references": [
+        "GUID:56255bd5d851a6243b63cb370cfc40b1",
+        "GUID:2238d58cada9b6746a548c6bff32c381"
```
joschemd-MS, line 6: Im assuming this is the MRTK.Theming reference? Is it now a hard dependency?

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
```diff
@@ -0,0 +1,178 @@
...
+                        themeItem.managedReferenceValue = new Theme.ThemeItem(themeDefinitionItemName, null);
+                    }
+                }
+
+                // Unified instantiation and auto-healing for missing or mismatched data types
+                Type expectedType = definitionItem.DataType?.Type;
+                SerializedProperty dataProp = themeItem.FindPropertyRelative(dataField);
+                Type actualType = dataProp?.managedReferenceValue?.GetType();
+
+                if (expectedType != null && expectedType != actualType && !failedTypes.Contains(expectedType))
```
joschemd-MS: NIT (Future consideration?):
The line/property 
private static readonly HashSet<Type> failedTypes = new HashSet<Type>();
is static and will exist over domain reloads, if the user attempts to fix something and doesn't compile after, they wont see the update until they do.

Recomendation: invalidate it during AssemblyReloadEvents.afterAssemblyReload.

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
```diff
@@ -0,0 +1,178 @@
...
+                {
+                    try
+                    {
+                        dataProp.managedReferenceValue = Activator.CreateInstance(expectedType);
```
joschemd-MS, line 176: NIT: for consistency (even though technically it shouldnt be in practice) dataProp can still be null here. Recomendation, extra explicit null check.

joschemd-MS (commented):
Let's goo! Almost there

org.mixedrealitytoolkit.uxcore/Theming/MRTK.UXCore.Theming.asmdef
keveleigh: Correct yeah, this is the MRTK.Theming reference.

image

BUT! It's not a hard dependency! Everything in here that depends on MRTK.Theming is gated on `#if MRTK_THEMING_PRESENT` which is based on the package being installed.

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
keveleigh: Added a null check here (and some other places!)

org.mixedrealitytoolkit.theming/Editor/ThemeEditor.cs
keveleigh: Good call! Updated to clear both in OnEnable and during AssemblyReloadEvents.afterAssemblyReload.

org.mixedrealitytoolkit.uxcore/Theming/MRTK.UXCore.Theming.asmdef
joschemd-MS: It just might give wierd warnings, but you're correct. Let's address this if it becomes a sore spot.

joschemd-MS (approved):

PR #1123, opened 2026-06-02
Update state visualizer and add tests

keveleigh:
* Fixed bug where the interactable could be changed but the visualizer would stay subscribed to the old interactable.
* Fixed bug where the animator could be changed but the visualizer wouldn't update the `AnimationPlayableOutput`.
* Added tests for hot-swapping `Interactable` and `Animator` on `StatefulVisualizer`.

image

### DIFF
--- a/org.mixedrealitytoolkit.uxcore/StateVisualizer/StateVisualizer.cs
+++ b/org.mixedrealitytoolkit.uxcore/StateVisualizer/StateVisualizer.cs
@@ -569,6 +569,11 @@ namespace MixedReality.Toolkit.UX
         /// Adds the provided effect to the state with name <paramref name="stateName"/>.
         /// Creates the state if it doesn't exist.
         /// </summary>
+        /// <remarks>
+        /// If this method is called at runtime after the component has been initialized,
+        /// you must manually call <see cref="Rebuild"/> to regenerate the <see cref="PlayableGraph"/>
+        /// so the new effect's playables are properly evaluated and connected.
+        /// </remarks>
         /// <param name="stateName">The name of the state to add the effect to.</param>
         /// <param name="effect">The effect to add.</param>
         internal void AddEffect(string stateName, IEffect effect)
@@ -584,6 +589,11 @@ namespace MixedReality.Toolkit.UX
         /// <summary>
         /// Removes the specified effect from the state with name <paramref name="stateName"/>.
         /// </summary>
+        /// <remarks>
+        /// If this method is called at runtime after the component has been initialized,
+        /// you must manually call <see cref="Rebuild"/> to properly disconnect and destroy
+        /// the removed effect's playables from the <see cref="PlayableGraph"/>.
+        /// </remarks>
         /// <param name="stateName">The name of the state to remove the effect from.</param>
         /// <param name="effect">The effect to remove.</param>
         /// <returns><see langword="true"/> if the effect was removed, <see langword="false"/> otherwise.</returns>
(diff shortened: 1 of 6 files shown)


org.mixedrealitytoolkit.uxcore/StateVisualizer/StateVisualizer.cs
```diff
@@ -130,12 +140,25 @@ public StatefulInteractable Interactable
...
+            {
+                if (animator != value)
+                {
+                    animator = value;
+                    if (playableGraph.IsValid())
+                    {
+                        animationPlayableOutput.SetTarget(animator);
+                    }
+                }
+            }
```
joschemd-MS, line 164: Since StateVisualizer is essentially taking ownership, should we be checking the previous enabled/play state (if there is an active animator) and clean up before changing, at the very least set the old one to disabled?

org.mixedrealitytoolkit.uxcore/StateVisualizer/StateVisualizer.cs
```diff
@@ -259,10 +363,39 @@ protected virtual void Start()
...
+                playableGraph.Stop();
+                enabled = false;
+            }
+            else
+            {
+                animator.enabled = true;
+                playableGraph.Play();
+                enabled = true;
+                sleepTimer = keepAliveTime;
+            }
```
joschemd-MS, line 410: Be careful on this and make sure it's fully tested, this section has caused me much grief. There are cases where the wieghts have been changed but haven't  been pushed through to the Animator before it is put to sleep thus we get quirks. I recommend considering calling playableGraph.Evaluate() again just before the sleep is enabled.

org.mixedrealitytoolkit.uxcore/Tests/Runtime/StateVisualizerTests.cs
```diff
@@ -340,6 +340,149 @@ public IEnumerator TestAnimatorMissing()
 
             Assert.IsTrue(cube.GetComponent<Animator>() != null, "An animator wasn't automatically added when it was missing!");
         }
+
+        /// <summary>
+        /// Tests that setting a new Interactable safely unsubscribes from the old and subscribes to the new.
+        /// </summary>
+        [UnityTest]
+        public IEnumerator TestInteractableHotSwap()
```
joschemd-MS, line 370: Recomendation: Add a minimal TestMixableEffect : IAnimationMixableEffect, IPlayableEffect to the test fixture (an AnimationClipPlayable with a one‑frame recognisable curve) and use it in the animator hot‑swap

keveleigh: Ah good idea! I added this to test against a specific value

org.mixedrealitytoolkit.uxcore/StateVisualizer/StateVisualizer.cs
keveleigh: Ah hmm...I wrote a few more tests, but I'll keep thinking on this one to make sure I've caught everything. "Good" to know it's been a troublesome spot, at least 😅

org.mixedrealitytoolkit.uxcore/StateVisualizer/StateVisualizer.cs
keveleigh: Thanks for catching this! I updated the setter to disable the old animator and ensure the new one's state matches the playable graph.

org.mixedrealitytoolkit.uxcore/StateVisualizer/StateVisualizer.cs
joschemd-MS: Thank you!

joschemd-MS (approved):

